PR #49, opened 2023-03-28
added library to mutate the kptfile

henderiw:
This PR contains a library that is used to mutate the kptfile.
- Marshal/UnMarshal and ParseObject operations
- Condition handling


Makefile
```diff
@@ -0,0 +1,11 @@
+.PHONY: fmt
+fmt: ## Run go fmt against code.
```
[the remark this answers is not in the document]
henderiw: the linter should be added in the repo not?

liamfallon (approved):
Just reviewing the code as a mechanism for learning, one tiny comment

ganchandrasekaran (approved):

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+type kptFile struct {
+	raw     []byte
+	kptfile *kptv1.KptFile
+}
+
+// Unmarshal decodes the raw document within the in byte slice and assigns decoded values into the out value.
+// it leverages the  "sigs.k8s.io/yaml" library
+func (r *kptFile) UnMarshal() (*kptv1.KptFile, error) {
```
gvbalaji, line 67: I am tempted to remove this  logic from this method and make it part of the NewMutator for the following reasons:
1. It will construct the structure completely during initialization and hence makes sure it's usable once it is constructed.
2. Without kptFile object this structure is not usable anyway.

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,167 @@
...
+	// returns an emoty slice
+	GetConditions() []kptv1.Condition
+}
+
+// NewMutator creates a new mutator for the kptfile
+// It expects a raw byte slice as input representing the serialized yaml file
+func NewMutator(b string) KptFile {
+	return &kptFile{
+		raw: []byte(b),
+	}
```
gvbalaji, line 57: My suggestion is to create the complete  object here by marshaling the byte array and make the state of the object usable immedietly

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	kptv1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
+	"sigs.k8s.io/yaml"
+)
+
+type KptFile interface {
+	// Unmarshal decodes the raw document within the in byte slice and assigns decoded values into the out value.
+	// it leverages the  "sigs.k8s.io/yaml" library
+	UnMarshal() (*kptv1.KptFile, error)
```
gvbalaji, line 30: My understanding is  this is mainly used  for CRUD operations for conditions in kpt files. if that's right my first thought looking at the interface is to separate the concerns into two separate interfaces . One for  file operations ( Marshall, Unmarshall etc) and one for CRUD on conditions.  It will keep the interface scope narrow and clear without any leak of abstractions. But there may be good reason why you may combined them into the same interface and that's fine as well. Just wanted to point out my thoughts.

krm-functions/lib/kptfile/v1/kpt_test.go
```diff
@@ -0,0 +1,190 @@
...
+						}
+					*/
+					if diff := cmp.Diff(tc.want[idx], got); diff != "" {
+						t.Errorf("TestGetCondition: -want, +got:\n%s", diff)
+					}
+				}
+			}
+		})
+	}
+}
```
gvbalaji, line 190: Tests are missing for other methods such as DeleteCondition, Unmarshall etc.

gvbalaji (changes requested):

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	// for each nex condition check if the kind
+	for _, new := range c {
+		exists := false
+		for i, existing := range r.GetKptFile().Status.Conditions {
+			if existing.Type != new.Type {
+				continue
+			}
+			r.GetKptFile().Status.Conditions[i] = new
```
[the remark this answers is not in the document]
henderiw: I added a comment for now to clarify what is needed as the fn would only be used once.

krm-functions/lib/kptfile/v1/condition_test.go
```diff
@@ -0,0 +1,79 @@
...
+				name:       "c",
+			},
+			want: "b.c",
+		},
+		{
+			input: object{
+				name: "c",
+			},
+			want: "c",
+		},
```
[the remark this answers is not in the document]
henderiw: added the test.

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,167 @@
...
+
+// DeleteCondition deletes the condition equal to the conditionType if it exists
+func (r *kptFile) DeleteCondition(ct string) {
+	if r.GetKptFile().Status == nil || len(r.GetKptFile().Status.Conditions) == 0 {
+		return
+	}
+
+	for idx, c := range r.GetKptFile().Status.Conditions {
+		if c.Type == ct {
+			r.GetKptFile().Status.Conditions = append(r.GetKptFile().Status.Conditions[:idx], r.GetKptFile().Status.Conditions[idx+1:]...)
```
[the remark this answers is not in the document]
henderiw: I believe go takes care. I never experienced issues with this so far. My understanding is that if the last element is deleted the last slice is empty so it returns the first part only.

https://www.geeksforgeeks.org/delete-elements-in-a-slice-in-golang/

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// GetKptFile returns the Kptfile as a go struct
+func (r *kptFile) GetKptFile() *kptv1.KptFile {
+	return r.kptfile
+}
+
+// SetConditions sets the conditions in the kptfile. It either updates the entry if it exists
+// or appends the entry if it does not exist.
+func (r *kptFile) SetConditions(c ...kptv1.Condition) {
```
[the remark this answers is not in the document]
henderiw: fn are not multithreaded but I agree it is better to have them than not to have them if someone uses it in a different context. Done

krm-functions/lib/kptfile/v1/kpt.go
henderiw: I agree and came to similar conclusion. Will change for the others as well.
Changed

krm-functions/lib/kptfile/v1/kpt.go
henderiw: I changed the name to New here as the code also has the namespace.
I did not like the NewMUtator name, but here is the challenge. For this library we only operate on kptfile and add/delete conditions and access information of the resource.

Now for ip allocation we have 2 different operations. 1 we have to create an allocation from scratch and 2 access and modify the data. I believe there need to be 2 different structs that handle this.

So I believe the one that create it I want to call NewGenerator and the other one New. So in IP Allocation we need both. A generator and a reader/parser and updater. This is my dilemma in general. For this we are ok with just the reader/parser and updater

krm-functions/lib/kptfile/v1/kpt.go
henderiw: Yes what I saw as a pattern is that these methods are coming back all the time.

// Marshal serializes the value provided into a YAML document based on "sigs.k8s.io/yaml".
	// The structure of the generated document will reflect the structure of the value itself.
	Marshal() ([]byte, error)
	// ParseKubeObject returns a fn sdk KubeObject; if something failed an error
	// is returned
	ParseKubeObject() (*fn.KubeObject, error)

So I am going to move this into a general Parser interface which is generic across all of the libraries.

so we get this.

type KptFile interface {
        // generic parser interface which contains Marshal/ParseKubeObject
	parser.Parser
	// GetKptFile returns the Kptfile as a go struct
	GetKptFile() *kptv1.KptFile
	// SetConditions sets the conditions in the kptfile. It either updates the entry if it exists
	// or appends the entry if it does not exist.
	SetConditions(...kptv1.Condition)
	// DeleteCondition deletes the condition equal to the conditionType if it exists
	DeleteCondition(ct string)
	// GetCondition returns the condition for the given ConditionType if it exists,
	// otherwise returns nil
	GetCondition(ct string) *kptv1.Condition
	// GetConditions returns all the conditions in the kptfile. if not initialized it
	// returns an emoty slice
	GetConditions() []kptv1.Condition
}

krm-functions/lib/kptfile/v1/kpt_test.go
henderiw: added the tests for these. DeleteCondition (here you see if I delete the last one it does not matter, so the procedure with append is safe).

80% coverage now

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,167 @@
...
+	// validate is the status is set, if not initialize the condition slice
+	if r.GetKptFile().Status == nil {
```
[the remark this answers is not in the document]
henderiw: it is done by kpt but I have added an additional check to ensure it is done.

krm-functions/lib/kptfile/v1/kpt.go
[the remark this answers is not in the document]
henderiw: I added some tests for this and confirms my understanding

krm-functions/lib/kptfile/v1/kpt.go
henderiw: I created this PR. So if we approve this I can update this 
https://github.com/nephio-project/nephio/pull/84

this PR specifies the Parser interface

Makefile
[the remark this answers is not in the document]
vjayaramrh: FYI I have added support for linter in PR at https://github.com/nephio-project/nephio/pull/83

liamfallon (approved):

krm-functions/lib/kptfile/v1/condition.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		sb.WriteString(o.Name)
+	}
+	return sb.String()
+}
+
+// GetGVKNFromConditionType return a KRM ObjectReference from a string
+// It expects an APIVersion with a / as a.b/c and a kind + name
+// if not it retruns an empty ObjectReference
+func GetGVKNFromConditionType(ct string) (o *corev1.ObjectReference) {
```
johnbelamaric, line 57: Do we need this reverse mapping? If we are using it in code, it locks us in pretty tightly to that naming schema. I can see it might be needed, for example, if we need to clean up stale conditions. Hmm.

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,170 @@
...
+	// GetConditions returns all the conditions in the kptfile. if not initialized it
+	// returns an emoty slice
+	GetConditions() []kptv1.Condition
+}
+
+// New creates a new parser interface
+// It expects a raw byte slice as input representing the serialized yaml file
+func New(b string) (KptFile, error) {
+	k := &kptv1.KptFile{}
+	if err := yaml.Unmarshal([]byte(b), k); err != nil {
```
johnbelamaric, line 54: I'm guessing any comments in the Kptfile get lost here?

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, err
+	}
+	return &kptFile{
+		kptfile: k,
+	}, nil
+}
+
+type kptFile struct {
+	m       sync.RWMutex
```
johnbelamaric, line 63: I am not sure the mutex is going to do what you hope. When running in a function, we are typically single-threaded anyway.

Are you intending this library to *also* be used by the specializer controllers? That is the concurrency we need to protect against - multiple specializers tweaking conditions at the same time. Unless all specializers run in the same binary (and we use leader election for that binary), we can't protect from that in this library. Instead, that has to be protected by Porch.

Consider:
 - controller A executes, reads the package
 - controller B executes, reads the package
 - controller A does its thing, and updates the conditions 
 - controller B does its thing, and updates the conditions - oops, it just overwrote controller A's conditions

Generally K8s uses optimistic concurrency. So, when the update is executed, it includes the original object revision; the server kicks it back if the object has since been updated. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency.

Ultimately, we need something like that. We may be able to get away with less in R1, though, but it will have race conditions.

henderiw: the goal was not to solve the concurrency issue we have, but rather having a way to protect if ever needed. I dot believe it harms us either, but I agree it should be solved by porch.

krm-functions/lib/kptfile/v1/condition.go
henderiw: GetConditionType does corev1.ObjectReference -> string
GetGVKNFromConditionType does string -> corev1.ObjectReference

Maybe the naming should be more aligned?

krm-functions/lib/kptfile/v1/kpt.go
johnbelamaric: Yeah. In fact in R1, the constraints I list above are probably true, so this will help in that case.

krm-functions/lib/kptfile/v1/kpt.go
henderiw: Good Question. I did a quick test and I don't see comments getting lost when unmarshaling using this method.

krm-functions/lib/kptfile/v1/condition.go
johnbelamaric: I understood that, I think the name is fine. My question was more about whether we need it. My philosophy is generally not to add any code until we need it. One less thing to test and maintain, if it's not being used.

henderiw: Here is how it is used in the fn-runtime.

// collect all conditions
		for _, c := range kf.GetConditions() {
			objRef := *kptfilelibv1.GetGVKNFromConditionType(c.Type)
			if len(r.cfg.Owns) == 0 {
				// collect all conditions for a wildcard
				r.inv.AddCondition(Wildcard, &objRef, &c)
			} else {
				for ref := range r.cfg.Owns {
					if strings.Contains(c.Type, kptfilelibv1.GetConditionType(&ref)) {
						r.inv.AddCondition(objRef.Name, &objRef, &c)
					}
				}
			}
		}
		
1. We track the conditions from the kptfile. E.g. a user can delete them and we want to ensure we are always aligned in an idempotent way to get back to the proper state.
2. we track the condition to tell what work is to be done and we use corev1.Objectoreference to store the data in a map to make lookup easier.

gvbalaji (changes requested):
Thanks a lot wim for addressing all the comments. Added a single review comment.

krm-functions/lib/kptfile/v1/kpt_test.go
```diff
@@ -0,0 +1,332 @@
...
+package v1
+
+import (
+	"testing"
+
+	kptv1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestParseObject(t *testing.T) {
```
vjayaramrh, line 26: usually the name should be "Test" +"name of function", i.e in this case it would be "Test"+"ParseKubeObject" => TestParseKubeObject

henderiw: fixed

krm-functions/lib/kptfile/v1/kpt.go
```diff
@@ -0,0 +1,170 @@
...
+	kptfile *kptv1.KptFile
+}
+
+// Marshal serializes the value provided into a YAML document based on "sigs.k8s.io/yaml".
+// The structure of the generated document will reflect the structure of the value itself.
+func (r *kptFile) Marshal() ([]byte, error) {
+	if r.kptfile == nil {
```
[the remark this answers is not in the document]
henderiw: fixed

krm-functions/lib/kptfile/v1/kpt.go
henderiw: this is the outcome of the test I did.

from:

apiVersion: kpt.dev/v1
kind: Kptfile
# comment A
metadata:
  name: pkg-upf
  #commentB
  annotations:
    config.kubernetes.io/local-config: "true"
info:
  description: upf package example
pipeline: {}


apiVersion: kpt.dev/v1
kind: Kptfile
# comment A
metadata:
  name: pkg-upf
  #commentB
  annotations:
    config.kubernetes.io/local-config: "true"
info:
  description: upf package example
pipeline: {}
status:
  conditions:
  - message: create new resource
    reason: req.nephio.org/v1alpha1.Interface.n3
    status: "False"
    type: ipam.nephio.org/v1alpha1.IPAllocation.n3
  - message: create new resource
    reason: req.nephio.org/v1alpha1.Interface.n3
    status: "False"
    type: k8s.cni.cncf.io/v1.NetworkAttachmentDefinition.n3
  - message: create new resource
    reason: req.nephio.org/v1alpha1.Interface.n4
    status: "False"
    type: ipam.nephio.org/v1alpha1.IPAllocation.n4
  - message: create new resource
    reason: req.nephio.org/v1alpha1.Interface.n4
    status: "False"
    type: k8s.cni.cncf.io/v1.NetworkAttachmentDefinition.n4
  - message: create new resource
    reason: req.nephio.org/v1alpha1.Interface.n6
    status: "False"
    type: ipam.nephio.org/v1alpha1.IPAllocation.n6
  - message: create new resource
    reason: req.nephio.org/v1alpha1.Interface.n6
    status: "False"
    type: k8s.cni.cncf.io/v1.NetworkAttachmentDefinition.n6

krm-functions/lib/kptfile/v1/kpt_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+package v1
+
+import (
+	"testing"
+
+	kptv1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
+	"github.com/google/go-cmp/cmp"
+)
+
```
vjayaramrh, line 25: The file is missing tests for the New() function. Something like below would help test the New() function

```
func TestNew(t *testing.T) {
	cases := map[string]struct {
		testInput        string
		errExpected      bool
		kptfileGenerated bool
	}{
		"NewSucceeds": {
			`apiVersion:`, false, true,
		},
		"NewFails": {
			`apiVersion`, true, false,
		},
	}
	for name, tc := range cases {
		t.Run(name, func(t *testing.T) {
			ret, e := New(tc.testInput)
			t.Log(ret, e)
			if tc.errExpected {
				assert.Error(t, e)
			} else {
				assert.NoError(t, e)
			}
			if tc.kptfileGenerated {
				assert.NotNil(t, ret)
			} else {
				assert.Nil(t, ret)
			}
		})
	}
}
```

krm-functions/lib/kptfile/v1/condition.go
johnbelamaric: Ok, thanks.

PR #64, opened 2023-03-30
kpt fn interface library

henderiw:
The interface library provides common functions that will be used by the krm-functions.


krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,83 @@
...
+	// The structure of the generated document will reflect the structure of the value itself.
+	Marshal() ([]byte, error)
+	// ParseKubeObject returns a fn sdk KubeObject; if something failed an error
+	// is returned
+	ParseKubeObject() (*fn.KubeObject, error)
+}
+
+// NewMutator creates a new mutator for the interface
+// It expects a raw byte slice as input representing the serialized yaml file
+func NewMutator(b string) Interface {
```
gvbalaji, line 41: Couple of comments:
1. Please consider renaming the function name to something like NewInterface or something like that which  is more isemantic.
2. Please consider deserializing the object stream iin this method itself which will create a complete usable object without any side effects.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+}
+
+type itfce struct {
+	raw   []byte
+	itfce *nephioreqv1alpha1.Interface
+}
+
+func (r *itfce) UnMarshal() (*nephioreqv1alpha1.Interface, error) {
+	i := &nephioreqv1alpha1.Interface{}
```
gvbalaji, line 53: Please see the comment above for the New method

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,83 @@
...
+// The structure of the generated document will reflect the structure of the value itself.
+func (r *itfce) Marshal() ([]byte, error) {
+	if r.itfce == nil {
+		return nil, errors.New("cannot marshal unitialized interface")
+	}
+	b, err := yaml.Marshal(r.itfce)
+	if err != nil {
+		return nil, err
+	}
+	r.raw = b
```
gvbalaji, line 71: During the init method we are assigning the raw byte stream. So what are we trying to do in this method?  What is the scenario in which there will be no value set for r.raw as it is happening during the initialization? In my view we should set both the raw and itfce during the init call and just return the values from these methods, more like getters.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,83 @@
...
+	return b, err
+}
+
+// ParseKubeObject returns a fn sdk KubeObject; if something failed an error
+// is returned
+func (r *itfce) ParseKubeObject() (*fn.KubeObject, error) {
+	b, err := r.Marshal()
```
gvbalaji, line 78: AGain not sure why need to call Marshall here.

krm-functions/lib/interface/v1alpha1/interface_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			got := tc.fn(itfce)
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Errorf("TestInterface: -want, +got:\n%s", diff)
+			}
+		})
+	}
+}
```
gvbalaji, line 79: Missing tests for ParseKubeObject.

gvbalaji (changes requested):
Thanks a lot for the PR.  Added review comments.

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: I changed to New as discussed in the kptfile PR. I add the unmarshal under the New so it is. done during init.
Fixed

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: fixed as per above

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: parse object only expects a byte. I reworked this a bit so you see raw is no longer there and we just get a go struct and we operate on the go struct only. Hope this clarifies

krm-functions/lib/interface/v1alpha1/interface_test.go
henderiw: added the missing tests

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,73 @@
...
+type Interface interface {
+    // ParseKubeObject returns a fn sdk KubeObject; if something failed an error
+    // is returned
+    ParseKubeObject() (*fn.KubeObject, error)
+    // GetInterface returns the interface as a go struct
+    GetInterface() *nephioreqv1alpha1.Interface
+}
+
+// New creates a new parser interface
+// It expects a raw byte slice as input representing the serialized yaml file
```
johnbelamaric, line 36: input type is string not `[]byte`

krm-functions/lib/interface/v1alpha1/interface_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+    "testing"
+
+    "github.com/google/go-cmp/cmp"
+    nephioreqv1alpha1 "github.com/nephio-project/api/nf_requirements/v1alpha1"
+)
+
+var itface = `apiVersion: req.nephio.org/v1alpha1
+kind: Interface
```
johnbelamaric, line 27: I am 90% sure that the way the code is written, we will lose comments in the YAML. Can you add a test here with comments, that goes through a read, edit, write cycle?

If I am right, then the pattern we can use instead os to:
 - on New, parse the thing into a KubeObject and cache that, it will be our internal storage format
 - GetInterface then would always then do KO -> YAML -> Interface, return that
 - Add a "SetInterface" that goes through all fields in the interface and does the corresponding KubeObject manipulations to set them in our internal storage
 
I suspect with `reflect` we could write helper libraries that do all the hard work *once* for all types.

go.sum
johnbelamaric, line 1: Do we need to update these dependencies as part of this PR? That is, are specific new versions needed? If not, I would prefer we do version updates of dependencies as a separate PR.

johnbelamaric (changes requested):

johnbelamaric:
Failure is: `krm-functions/lib/interface/v1alpha1/interface.go:20: File is not `gofmt`-ed with `-s` (gofmt)`

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type itfce struct {
+    itfce *nephioreqv1alpha1.Interface
+}
+
+// Marshal serializes the value provided into a YAML document based on "sigs.k8s.io/yaml".
+// The structure of the generated document will reflect the structure of the value itself.
+func (r *itfce) marshal() ([]byte, error) {
+    if r.itfce == nil {
```
vjayaramrh, line 54: question: Will `r.itfce` be ever nil unless someone intentionally set it to nil? We may not need this method `marshal` and could just use the `yaml.Marshal(r.itfce)` if it `r.itfce` is not nil.

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: fixed

krm-functions/lib/interface/v1alpha1/interface_test.go
henderiw: ok I changed the strategy accordingly

go.sum
henderiw: the api repo requires these.

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: the strategy changed aligned with what John was proposing. Now I added a protection as the user could set it to nil.

henderiw:
I have checked in a new approach. this would be used for all fn: generators, mutators, etc.
@johnbelamaric @gvbalaji @vjayaramrh 

We store local all in kubeObject and have some methods around it.

The New would create the interface. You can either supply a go struct or a raw byte which would be yaml. 

We have 3 operations to perform on these objects.

- initialise the interface:
-> when you generate. We could use New(go struct)
-> any other scenario. We use the []byte

Afterwards we need to:
- access data: we could standardise on the go struct and leverage the Get methods it exposes. (this means no specific get methods are needed)
- mutate data: we have a gostruct to set the complete object and individual setter methods we need to write (or codeine going fwd)

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	if err := r.o.SetNestedField(s, "spec", "networkInstance", "name"); err != nil {
+		return err
+	}
+	return nil
+}
+
+// SetInterfaceSpec sets the spec attributes in the kubeobject
+func (r *itfce) SetInterfaceSpec(spec *nephioreqv1alpha1.InterfaceSpec) error {
+	if spec != nil {
```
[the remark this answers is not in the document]
henderiw: that is a good point I will delete the value when not set in the go struct.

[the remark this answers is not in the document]
henderiw: I added the delete methods and applied them to the spec if they are not present in the spec. At least the once that can be omitted.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -46,13 +49,29 @@ type Interface interface {
...
+
+	DeleteAttachmentType() error
+
+	DeleteCNIType() error
 }
 
 // New creates a new parser interface
 // It expects a raw byte slice as input representing the serialized yaml file
-func New(b []byte) (Interface, error) {
+func New(x any) (Interface, error) {
```
johnbelamaric, line 62: Alternatively you can have `NewFromYAML([]byte)` and `NewFromGoStruct(*nephioreqv1alpha1.Interface)`. Is there a reason to use the same name?

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -157,23 +206,33 @@ func (r *itfce) SetNetworkInstanceName(s string) error {
 	return nil
 }
 
-// SetInterfaceSpec sets the spec attributes in the kubeobject
-func (r *itfce) SetInterfaceSpec(spec *nephioreqv1alpha1.InterfaceSpec) error {
+// SetSpec sets the spec attributes in the kubeObject
+func (r *itfce) SetSpec(spec *nephioreqv1alpha1.InterfaceSpec) error {
```
johnbelamaric, line 210: Rather than nest everything in `if spec != nil` it's easier to read if you do:

```
 if spec == nil {
  return nil
}
```

Or maybe you want to return an error in that case? Either way - return early instead of nesting the whole body in an "if".

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: yes that works as well. I agree it is better to be explicit. Will change.

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: yes agreed will fix. Was just hacking something together to agree on the approach going fwd as this is applicable to all libraries.

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: fixed

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: fixed

johnbelamaric (commented):
Nice! One non-blocking comment, plus this: please add a test that proves comments are not lost in the YAML.

Thanks.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+	return nil
+}
+
+func (r *itfce) SetCNIType(s string) error {
+	if r.o == nil {
+		return fmt.Errorf("KubeObject not initialized")
+	}
+	// validation -> should be part of the interface api repo
+	switch s {
+	case string(nephioreqv1alpha1.CNITypeIPVLAN):
```
gvbalaji, line 173: It's tempting for me to store the supported CNI types in a slice (in a file where they are defined)  named SupportedCNIs and add a method  "IsCNISupported(att String) bool". That method will look up in the slice and return a boolean. That way we can capture this in a common way and future proof as well. All we need to do is update the slice when we add/delete support for an CNI type.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	_, err := r.o.RemoveNestedField("spec", "cniType")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *itfce) SetNetworkInstanceName(s string) error {
+	if r.o == nil {
```
gvbalaji, line 198: Similar to getters , this block of code is same for setters and can be refactored to a common method. Call the validate method and then call the common setter method.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+
+	if err := r.o.SetNestedField(s, "spec", "attachmentType"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *itfce) DeleteAttachmentType() error {
+	if r.o == nil {
+		return fmt.Errorf("KubeObject not initialized")
```
gvbalaji, line 158: Similar to Get, all the Deletes can be refactored 

func (r *itfce) DeleteAttachmentType() error {
     return deleteField("spec", "attachmentType")
}

func (r *itfce) deleteField(fields...string) error {
	if r.o == nil {
		return fmt.Errorf("KubeObject not initialized")
	}
	_, err := r.o.RemoveNestedField(fields...)
	if err != nil {
		return err
	}
	return nil
}

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+		return fmt.Errorf("unknown cniType")
+	}
+
+	if err := r.o.SetNestedField(s, "spec", "cniType"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *itfce) DeleteCNIType() error {
```
gvbalaji, line 186: Same refactor comment as above for the DeleteAttachment()

krm-functions/lib/interface/v1alpha1/interface_test.go
```diff
@@ -0,0 +1,440 @@
...
+		}
+
+		t.Run(name, func(t *testing.T) {
+			err := i.DeleteCNIType()
+			assert.NoError(t, err)
+
+			fmt.Println(i.GetKubeObject().String())
+		})
+	}
+}
```
gvbalaji, line 440: May be missing test for DeleteAttachmentType.

gvbalaji (changes requested):
Thanks @henderiw . I have added a few review comments mostly around refactoring for reuse and simplicity.

krm-functions/lib/interface/v1alpha1/interface_test.go
```diff
@@ -0,0 +1,440 @@
...
+
+var itfaceEmpty = `apiVersion: req.nephio.org/v1alpha1
+kind: Interface
+metadata:
+  name: n3
+  annotations:
+    config.kubernetes.io/local-config: "true"
+`
+
+func TestNew(t *testing.T) {
```
vjayaramrh, line 50: It maybe better to rename this as `TestNewFromYaml`

vjayaramrh, line 50: May need a test for `NewFromGoStruct`

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+type itfce struct {
+	o *fn.KubeObject
+}
+
+// GetKubeObject returns the present kubeObject
+func (r *itfce) GetKubeObject() *fn.KubeObject {
+	return r.o
+}
+
+func (r *itfce) GetGoStruct() (*nephioreqv1alpha1.Interface, error) {
```
vjayaramrh, line 88: This method would need a test function in `interface_test.go` file

krm-functions/lib/interface/v1alpha1/interface_test.go
henderiw: added

henderiw: changed

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: added

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+	x := &nephioreqv1alpha1.Interface{}
+	if err := yaml.Unmarshal([]byte(r.o.String()), x); err != nil {
+		return nil, err
+	}
+	return x, nil
+}
+
+func (r *itfce) GetAttachmentType() string {
+	if r.o == nil {
```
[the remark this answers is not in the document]
henderiw: great suggestion. Fixed

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+		return ""
+	}
+	if !ok {
+		return ""
+	}
+	return s
+}
+
+func (r *itfce) GetCNIType() string {
+	if r.o == nil {
```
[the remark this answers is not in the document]
henderiw: fixed

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+	DeleteAttachmentType() error
+
+	DeleteCNIType() error
+}
+
+// NewFromYAML creates a new parser interface
+// It expects a raw byte slice as input representing the serialized yaml file
+func NewFromYAML(b []byte) (Interface, error) {
+	o, err := fn.ParseKubeObject(b)
+	return &itfce{
```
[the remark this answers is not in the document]
henderiw: I believe the fn does this, but I made it explicit as you suggested

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+		return ""
+	}
+	if !ok {
+		return ""
+	}
+	return s
+}
+
+func (r *itfce) GetNetworkInstanceName() string {
+	if r.o == nil {
```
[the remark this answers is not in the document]
henderiw: fixed

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,238 @@
...
+		return ""
+	}
+	if !ok {
+		return ""
+	}
+	return s
+}
+
+func (r *itfce) SetAttachmentType(s string) error {
+	if r.o == nil {
```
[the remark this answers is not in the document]
henderiw: resolved

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: fixed

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: done

krm-functions/lib/interface/v1alpha1/interface_test.go
henderiw: added

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,240 @@
...
+	// DeleteAttachmentType deletes the attachmentType from the spec
+	DeleteAttachmentType() error
+	// DeleteAttachmentType deletes the attachmentType from the spec
+	DeleteCNIType() error
+}
+
+// NewFromYAML creates a new parser interface
+// It expects a raw byte slice as input representing the serialized yaml file
+func NewFromYAML(b []byte) (Interface, error) {
+	o, err := fn.ParseKubeObject(b)
```
vjayaramrh, line 69: question: Looking at the fn.ParseKubeObject(), it seems like `o` will never be `nil`. Is this assumption valid?

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	if err := r.o.SetNestedField(s, fields...); err != nil {
+		return err
+	}
+	return nil
+}
+
+// deleteNestedField is a generic utility function that deletes
+// a string slice representing the path from the yaml doc
+func (r *itfce) deleteNestedField(fields ...string) error {
```
vjayaramrh, line 231: I am of opinion that the `deleteNestedField` method can be removed and the calling function can directly invoke the `r.o.RemoveNestedField` instead by adding a check for `r.o != nil` at the top of the calling method.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -0,0 +1,240 @@
...
+	if !ok {
+		return ""
+	}
+	return s
+}
+
+// setNestedField is a generic utility function that sets a string on
+// a string slice representing the path in the yaml doc
+func (r *itfce) setNestedField(s string, fields ...string) error {
+	if r.o == nil {
```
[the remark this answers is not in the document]
henderiw: I cannot find a way to have r.o set to nil, but it would be nil it will panic the method.

krm-functions/lib/interface/v1alpha1/interface.go
gvbalaji: This is a helper(private) method used in other public methods as part of the refactor.

gvbalaji (approved):
I understand once the PR in the API repo is merged we will use validation routines from there. That can come as another PR and we can make changes before the PR.

henderiw:
ok thx. there is a few more changes to be added but we can do this through another PR

PR #65, opened 2023-03-30
added kpt resourcelist library

henderiw:
library to manipulate the resource list of a kit package.

type ResourceList interface {
	AddResult(err error, obj *fn.KubeObject)
	GetObjects() fn.KubeObjects
	SetObject(obj *fn.KubeObject)
	SetObjectWithDeleteTimestamp(obj *fn.KubeObject)
	AddObject(obj *fn.KubeObject)
	DeleteObject(obj *fn.KubeObject)
}


krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (r *resourceList) AddObject(obj *fn.KubeObject) {
+	r.rl.Items = append(r.rl.Items, obj)
+}
+
+func (r *resourceList) DeleteObject(obj *fn.KubeObject) {
+	for idx, o := range r.rl.Items {
+		if o.GetAPIVersion() == obj.GetAPIVersion() && o.GetKind() == obj.GetKind() && o.GetName() == obj.GetName() {
+			r.rl.Items = append(r.rl.Items[:idx], r.rl.Items[idx+1:]...)
```
gvbalaji, line 88: Basic question on the index range if deleting the last item in the slice. Does idx+1 cause any range errors?

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,91 @@
...
+			exists = true
+			break
+		}
+	}
+	if !exists {
+		r.AddObject(obj)
+	}
+}
+
+func (r *resourceList) SetObjectWithDeleteTimestamp(obj *fn.KubeObject) {
```
gvbalaji, line 67: It would be good to document what this method is doing.  Before assigning to "r.rl.Items[idx]" did the value of "obj" change?

krm-functions/lib/kptrl/kptrl_test.go
```diff
@@ -0,0 +1,19 @@
+/*
```
gvbalaji, line 1: I guess you would be adding tests later .

gvbalaji (changes requested):
Thanks for the PR. Added a few comments.

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (r *resourceList) GetObjects() fn.KubeObjects {
+	return r.rl.Items
+}
+
+func (r *resourceList) SetObject(obj *fn.KubeObject) {
+	exists := false
+	for idx, o := range r.rl.Items {
+		if o.GetAPIVersion() == obj.GetAPIVersion() && o.GetKind() == obj.GetKind() && o.GetName() == obj.GetName() {
```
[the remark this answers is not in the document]
henderiw: ok add IsGVKN fn.

krm-functions/lib/kptrl/kptrl.go
henderiw: go handles this properly

krm-functions/lib/kptrl/kptrl_test.go
henderiw: done. tests added with 100% coverage

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+// New creates a new ResourceList interface
+// concurrency is handles in the methids
+func New(rl *fn.ResourceList) ResourceList {
+	return &resourceList{
+		rl: rl,
+	}
+}
+
+type resourceList struct {
+	m  sync.RWMutex
```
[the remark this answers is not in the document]
henderiw: No but I added it as requested by Bala. I don't believe it harms either.

makefile
```diff
@@ -0,0 +1,53 @@
+GO_VERSION ?= 1.20.2
```
vjayaramrh, line 1: This file `makefile` can be removed as the `Makefile` has already been merged into the main branch

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
+/*
+ Copyright 2023 Nephio.
```
johnbelamaric, line 2: I think the "proper" form according to our OSPO would be "The Nephio Authors" or "The Nephio Project Authors".

@vjayaramrh @aravind254 @radoslawc Can we add something to Prow that checks these headers? Is that what is done in other projects?

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+	// SetObject sets the object in the resourcelist items. It either updates/overrides
+	// the entry if it exists or appends the entry if it does not exist in the resourcelist
+	// It uses APIVersion, Kind and Name to check the object uniqueness
+	SetObject(obj *fn.KubeObject)
+	// DeleteObject deletes the object from the resourcelist if it exists.
+	DeleteObject(obj *fn.KubeObject)
+}
+
+// New creates a new ResourceList interface
+// concurrency is handles in the methids
```
johnbelamaric, line 45: nit: "Concurrency is handled in the methods"

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+func New(rl *fn.ResourceList) ResourceList {
+	return &resourceList{
+		rl: rl,
+	}
+}
+
```
johnbelamaric, line 51: Can you add a comment like:

"Typically in functions there is no concurrency, but this library may be used in specializers running concurrently in the same binary, and may share the same resource list".

I am still somewhat doubtful that's the case, but at least it explains why this is here.

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+type resourceList struct {
+	m  sync.RWMutex
+	rl *fn.ResourceList
+}
+
+// AddResult add a result with error and corresponding kubeObject by
```
johnbelamaric, line 57: nits:
- s/kubeObject/KubeObject/
- s/add a/adds a/
- s/resourcelist/resourceList/

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+
+package kptrl
+
+import (
+	"sync"
+
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+type ResourceList interface {
```
johnbelamaric, line 25: Ignoring the concurrency controls (which I don't think are needed), this seems to mostly wrap fn.ResourceList and add a couple methods. 

I think it would be simpler and more idiomatic to instead do:

```
type ResourceList struct {
  fn.ResourceList
  ...add new methods here...
}
```

And just let folks access the underlying fields directly, rather than having Getters that just return the fiedl from the wrapped object.

That would eliminate these methods, because you would just access the underlying fields instead:
- AddResult
- GetResults
- GetObjects

So basically, what we would have in the end is a library that adds the GetObject, SetObject, DeleteObject convenience methods.

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+// GetObjects returns all items from the resourcelist
+func (r *resourceList) GetObjects() fn.KubeObjects {
+	r.m.RLock()
+	defer r.m.RUnlock()
+	return r.rl.Items
+}
+
+// SetObject sets the object in the resourcelist items. It either updates/overrides
+// the entry if it exists or appends the entry if it does not exist in the resourcelist
+// It uses APIVersion, Kind and Name to check the object uniqueness
```
johnbelamaric, line 94: We should also use Namespace; in the case where it's local-config and has no namespace, that should still work.

johnbelamaric (changes requested):

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"sync"
+
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+)
+
+type ResourceList interface {
+	// AddResult add a result with error and corresponding kubeObject by
+	// appending a result to the result slice in the resourcelist
+	AddResult(err error, obj *fn.KubeObject)
```
vjayaramrh, line 28: Trying to understand why `AddResult` takes only an error object? If we wanted only add error result, could we not use `ErrorResult` at https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/fn/result.go#L274

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+	GetResults() fn.Results
+	// GetObject return an fn sdk KubeObject by comparing the APIVersion, Kind and Name
+	// if the object is found the corresponding obj is returned, if not nil is returned
+	GetObject(obj *fn.KubeObject) *fn.KubeObject
+	// GetObjects returns all items from the resourcelist
+	GetObjects() fn.KubeObjects
+	// SetObject sets the object in the resourcelist items. It either updates/overrides
+	// the entry if it exists or appends the entry if it does not exist in the resourcelist
+	// It uses APIVersion, Kind and Name to check the object uniqueness
+	SetObject(obj *fn.KubeObject)
```
vjayaramrh, line 39: Would not the `UpsertObjectToItems` at https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/fn/resourcelist.go#L191 not provide the equivalent of `SetObject`?

krm-functions/lib/kptrl/kptrl.go
vjayaramrh: > nit: "Concurrency is handled in the methods"

Am wondering if add the misspell to the linter could help, 
@radoslawc FYI

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// SetObject sets the object in the resourcelist items. It either updates/overrides
+// the entry if it exists or appends the entry if it does not exist in the resourcelist
+// It uses APIVersion, Kind and Name to check the object uniqueness
+func (r *resourceList) SetObject(obj *fn.KubeObject) {
+	r.m.Lock()
+	defer r.m.Unlock()
+	exists := false
+	for idx, o := range r.rl.Items {
```
vjayaramrh, line 99: I was thinking if replacing L99 to L108 with the  `UpsertObjectToItems` at https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/go/fn/resourcelist.go#L191 could help.?

krm-functions/lib/kptrl/kptrl.go
vjayaramrh, line 53: Can we add scenario tests that demonstrates the need for the sync.RWMutex?

krm-functions/lib/kptrl/kptrl.go
gvbalaji: I had originally requested for thread safety is some of these libraries. My intention was not to make assumptions on how the library would be used and provide thread safety. I guess the cost is not too much and the callers don;t need to worry about the side effects if at all.

gvbalaji (commented):
Thanks @henderiw . Added a few comments and replied to a few comments as well.

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,133 @@
...
+	// if the object is found the corresponding obj is returned, if not nil is returned
+	GetObject(obj *fn.KubeObject) *fn.KubeObject
+	// GetObjects returns all items from the resourcelist
+	GetObjects() fn.KubeObjects
+	// SetObject sets the object in the resourcelist items. It either updates/overrides
+	// the entry if it exists or appends the entry if it does not exist in the resourcelist
+	// It uses APIVersion, Kind and Name to check the object uniqueness
+	SetObject(obj *fn.KubeObject)
+	// DeleteObject deletes the object from the resourcelist if it exists.
+	DeleteObject(obj *fn.KubeObject)
```
ganchandrasekaran, line 41: @henderiw Based on the latest fn-runtime library Some functions are missing;
e.g: AddObject(o)

Can you please update this to latest?

ganchandrasekaran: @henderiw Similarly `NewMutator` function is missing from the latest kptfile lib: https://github.com/nokia/nephio-nephio/tree/kptfile-lib

can you please add it.

krm-functions/lib/kptrl/kptrl.go
henderiw: fixed

krm-functions/lib/kptrl/kptrl.go
henderiw: added the note

krm-functions/lib/kptrl/kptrl.go
henderiw: fixed

krm-functions/lib/kptrl/kptrl.go
henderiw: fixed

krm-functions/lib/kptrl/kptrl.go
henderiw: added namespace

krm-functions/lib/kptrl/kptrl.go
henderiw: ErrorResult is to provide the overall result. here we are adding individual results that are seen when running the fn.

krm-functions/lib/kptrl/kptrl.go
henderiw: updated to the fn lib

krm-functions/lib/kptrl/kptrl.go
henderiw: updated to the fn lib

krm-functions/lib/kptrl/kptrl.go
henderiw: AddObject is set object and NewMutator is New

krm-functions/lib/kptrl/kptrl.go
henderiw: fn.ResourceList is not an interface. What we could do is add the methods the resourceList offers and these to the interface?

henderiw: Eventually all of this should end up in the kpt fn sdk in my view.

krm-functions/lib/kptrl/kptrl.go
henderiw: done

krm-functions/lib/kptrl/kptrl.go
johnbelamaric: Sorry, right, you can't add new methods there. My point would have been to use a struct and just allow access to the underlying fn.ResourceList and then only implement the new functions that do extra stuff on top of that.

I still don't think the mutex is needed based on how these objects are actually used, and while it doesn't add significant runtime overhead, it will raise questions (like it has for me).

Beyond that, most of the methods are single line - except GetObject and DeleteObject, I think. So, apart from those two, does it make sense to add another layer folks have to learn on top of the SDK?

gvbalaji: If thread safety is not needed, agree to the point that most of the methods can be directed  to the underlying rl object.

makefile
henderiw: resolved

krm-functions/lib/kptrl/kptrl.go
henderiw: ok I removed the thread safety and add a struct that holds the rl as John suggested. Let me know if this is ok

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -17,65 +17,33 @@
...
 	return r.rl.Results
 }
 
 // GetObject return an fn sdk KubeObject by comparing the APIVersion, Kind, Name and Namespace
 // if the object is found the corresponding obj is returned, if not nil is returned
-func (r *resourceList) GetObject(obj *fn.KubeObject) *fn.KubeObject {
-	r.m.RLock()
-	defer r.m.RUnlock()
+func (r *ResourceList) GetObject(obj *fn.KubeObject) *fn.KubeObject {
 	for _, o := range r.rl.Items {
```
johnbelamaric, line 47: Just noticed this: https://pkg.go.dev/github.com/GoogleContainerTools/kpt-functions-sdk/go/fn#KubeObjects.Where

This, along with change described above, would then become:

```

func (r *ResourceList) GetObject(obj *fn.KubeObject) *fn.KubeObject {
        return r.Items.Where(func(ko *KubeObject) bool { isGVKNNEqual(ko, obj) })
}

```

johnbelamaric (changes requested):
Sorry, I don't mean to be a pain. I just want to make sure we're not adding code that changes the interface but otherwise doesn't add convenience.

Accessing the underlying, exported fields is standard practice in Go, rather than using getters. The `fn` sdk provides a lot of accessor methods, which look to be redundant with these, once we make the exported fields available - or more to the point, in this case we create an extension and we can re-use the fields and methods of the original.

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -17,65 +17,33 @@
...
-	// GetObjects returns all items from the resourceList
-	GetObjects() fn.KubeObjects
-	// SetObject sets the object in the resourceList items. It either updates/overrides
-	// the entry if it exists or appends the entry if it does not exist in the resourceList
-	// It uses APIVersion, Kind, Name and Namespace to check the object uniqueness
-	SetObject(obj *fn.KubeObject) error
-	// DeleteObject deletes the object from the resourceList if it exists.
-	DeleteObject(obj *fn.KubeObject)
+type ResourceList struct {
+	rl *fn.ResourceList
```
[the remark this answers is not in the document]
henderiw: fixed.

krm-functions/lib/kptrl/kptrl.go
henderiw: where returns multiple objects AFAIS. So wanted to return 1 but will return all that match and use this

johnbelamaric: Ah, right

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,77 @@
...
+
+// GetResults gets the results slice from the resourceList
+func (r *ResourceList) GetResults() fn.Results {
+	return r.Results
+}
+
+// GetObjects return an fn sdk KubeObject by comparing the APIVersion, Kind, Name and Namespace
+// if the object is found the corresponding obj is returned, if not nil is returned
+func (r *ResourceList) GetObjects(obj *fn.KubeObject) fn.KubeObjects {
+	return r.Items.Where(func(ko *fn.KubeObject) bool { return isGVKNNEqual(ko, obj) })
```
johnbelamaric, line 41: I think if you want a single object to be returned that's fine, my suggestion to use Where wasn't intended to change the function signature.

henderiw: it is ok like this. I just pick the first one

krm-functions/lib/kptrl/kptrl.go
```diff
@@ -0,0 +1,77 @@
...
+
+// GetObjects returns all items from the resourceList
+func (r *ResourceList) GetAllObjects() fn.KubeObjects {
+	return r.Items
+}
+
+// SetObject sets the object in the resourceList items. It either updates/overrides
+// the entry if it exists or appends the entry if it does not exist in the resourceList
+// It uses APIVersion, Kind, Name and Namespace to check the object uniqueness
+func (r *ResourceList) SetObject(obj *fn.KubeObject) error {
```
gvbalaji, line 52: We can directly say
 return r.UpsertObjectToItems(obj, nil, true)

krm-functions/lib/kptrl/kptrl_test.go
```diff
@@ -0,0 +1,342 @@
...
+			got := r.GetObjects(tObj)
+
+			if got != nil {
+				t.Errorf("TestDeleteObject: -want: nil, +got:%v\n", got)
+			}
+		})
+	}
+}
+
+func TestConurrency(t *testing.T) {
```
gvbalaji, line 300: This is not needed now as we have removed the mutex.

gvbalaji (changes requested):

krm-functions/lib/kptrl/kptrl.go
henderiw: agreed and fixed

krm-functions/lib/kptrl/kptrl_test.go
henderiw: agreed and removed

gvbalaji (approved):
looks good to me.

PR #67, opened 2023-03-30
added nad-lib

henderiw:
nad-lib helps the and for fn operations on NAD resources


gvbalaji (commented):
looks good so far. Thanks.

krm-functions/lib/nad/v1/nad_test.go
```diff
@@ -0,0 +1,121 @@
+/*
+Copyright 2023 Nephio.
+Copyright 2023 Samsung.
```
johnbelamaric, line 3: You should check with your internal team if this is needed, or if "Nephio Project Authors" is sufficient. At least Google does not require this for our contributions; apparently these headers are actually not particularly meaningful for legal reasons (that is not legal advice...). The administrative overhead of carrying a lot of these notices from every contributor is quite painful; thus, changing it all to a generic "project authors" and using the Git history is preferred. For example, see: https://github.com/GoogleContainerTools/kpt/pull/3907

Now, if we are pulling this file from something that someone at Samsung wrote, and they have this header, and it is not someone at Samsung doing the commit, then we have to keep the header, according to the Apache license. That is, only someone from Samsung could remove this.

Sicne Wim is doing the commit, we may need it. But If Ganesh is doing the commit then it wouldn't be needed - assuming Wim copied this from something Ganesh wrote somewhere else.

johnbelamaric (dismissed):
I think this should be reconsidered along the same lines as my comments in the other library PRs.

krm-functions/lib/nad/v1/nad_test.go
ganchandrasekaran: Hi John, I too think having "Nephio Project Authors" is sufficient. I will also check this internally with my organisation. 
This practise came from contribution towards ONAP, where anyone updating a file will append their organisation’s Copyright on top. Please note in ONAP, there is no concept of PR, it was all direct review and commit. But in Nephio, a PR can have many commits from multiple authors. In summary, I agree with your legal reasoning and happy to remove the same from this file. We will just kept Nephio’s Copyright on top. Thanks.

krm-functions/lib/nad/v1/nad_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestParseKubeObjectByNADInterface(t *testing.T) {
```
vjayaramrh, line 28: Maybe a good idea to come the two different test functions into one single `TestParseKubeObject` using table/map driven tests

ganchandrasekaran:
Please note: `github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1` defines the library for NetworkAttachmentDefinition. On the same the Config is defined as a String

```type NetworkAttachmentDefinitionSpec struct {
	Config string `json:"config"`
}```

In practice this Config is a JSON object and it can be configured as several types such as a VLAN, with IP Address information and with a static IP address. Specific to Nephio I see the following three scenarios:

Case-1: Only VLAN and no IPAM
```
spec:
  config: '{
      "cniVersion": "0.3.0",
      "type": "sriov",
      "vlan": 2000
    }'
```

Case-2: Only IPAM and no VLAN (in IPAM only static is supported)
```
  spec:
    config: '{
  "cniVersion": "0.3.1",
  "plugins": [
    {
      "type": "sriov",
      "capabilities": {
        "ips": true
      },
      "master": "eth1",
      "mode": "bridge",
      "ipam": {
        "type": "static",
        "addresses": [
          {
            "address": "14.0.0.2/24",
            "gateway": "14.0.0.1"
          }
        ]
      }
    }
  ]
}'
 ```

Case-2: Both IPAM and VLAN is present

```
  spec:
    config: '{
  "cniVersion": "0.3.1",
  "vlan": 2000
  "plugins": [
    {
      "type": "sriov",
      "capabilities": {
        "ips": true
      },
      "master": "eth1",
      "mode": "bridge",
      "ipam": {
        "type": "static",
        "addresses": [
          {
            "address": "14.0.0.2/24",
            "gateway": "14.0.0.1"
          }
        ]
      }
    }
  ]
}'
 ```

File: krm-functions/lib/nad/v1/nad.go [hunk 3]
```diff
@@ -0,0 +1,283 @@
...
+
+func (r *NadStruct) SetCNIType(cniType string) error {
+	if cniType != "" {
+		nadConfigStruct := r.GetNadConfig()
+		nadConfigStruct.Plugins[0].Type = cniType
+		b, err := json.Marshal(nadConfigStruct)
+		if err != nil {
+			return err
+		}
+		return r.K.SetNestedString(string(b), ConfigType...)
```
kispaljr, line 229: this code is repeated in multiple setters, so I would refactor it to a new method, like:
```
func (r *NadStruct) SetNadConfig(config NadConfig) error {
	b, err := json.Marshal(config)
	if err != nil {
		return err
	}
	return r.K.SetNestedString(string(b), ConfigType...)
}
```
and called it here

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		return err
+	}
+	return r.K.SetNestedString(string(b), ConfigType...)
+}
+
+func (r *NadStruct) SetCNIType(cniType string) error {
+	if cniType != "" {
+		nadConfigStruct := r.GetNadConfig()
+		nadConfigStruct.Plugins[0].Type = cniType
```
kispaljr, line 224: check if `len(.....Plugins) > 0` before getting the first element

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (r *NadStruct) SetNadMaster(nadMaster string) error {
+	if nadMaster != "" {
+		nadConfigStruct := r.GetNadConfig()
+		nadConfigStruct.Plugins[0].Master = nadMaster
+		b, err := json.Marshal(nadConfigStruct)
+		if err != nil {
+			return err
+		}
+		return r.K.SetNestedString(string(b), ConfigType...)
```
kispaljr, line 257: call `r.SetNadConfig(nadConfigStruct)` instead of repeating code

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+	} else {
+		return fmt.Errorf("unknown IPAM address")
+	}
+}
+
+func (n *NadConfig) ToString() string {
```
kispaljr, line 277: Consider renaming this to String() to comply with Stringer interface: https://go.dev/tour/methods/17

kispaljr (changes requested):
I like the approach of defining the go structs, parse Config properly and do the manipulation on the go structs.

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+// It expects a go struct representing the interface krm resource
+func NewFromGoStruct(x *nadv1.NetworkAttachmentDefinition) (*NadStruct, error) {
+	p, err := kubeobject.NewFromGoStruct[*nadv1.NetworkAttachmentDefinition](x)
+	if err != nil {
+		return nil, err
+	}
+	return &NadStruct{K: *p}, nil
+}
+
+func (r *NadStruct) GetCniSpecType(fields ...string) CniSpecType {
```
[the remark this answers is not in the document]
kispaljr: I forgot to mention that similarly `SetCniSpecType(cniType CniSpecType)` also doesn't seem to be useful

krm-functions/lib/nad/v1/nad.go
kispaljr: If GetNadConfig() would guarantee that Plugins is not empty this would be fine (I would still write this fact a comment before this line), but right now it doesn't give you that guarantee (and probably it is good this way, and it shouldn't)

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+		return r.K.SetNestedString(string(b), ConfigType...)
+	} else {
+		return fmt.Errorf("unknown IPAM address")
+	}
+}
+
+func (n *NadConfig) ToString() string {
+	b, err := json.Marshal(n)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
kispaljr: you know what, I changed my mind. Panicking here is better than returning with anything, so ignore my comment above :)

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+type CniSpecType int64
+
+const (
+	Both     CniSpecType = iota // 0
+	IpamType                    // 1
+	VlanType                    // 2
+)
+
+type NadStruct struct {
+	K kubeobject.KubeObjectExt[*nadv1.NetworkAttachmentDefinition]
```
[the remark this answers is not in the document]
ganchandrasekaran: This is in to align with ipalloc and interface lib.. 
e.g. https://github.com/nephio-project/nephio/blob/main/krm-functions/lib/ipalloc/v1alpha1/ipalloc.go#L26

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+	C CniSpecType
+}
+
+// NewFromKubeObject creates a new parser interface
+// It expects a *fn.KubeObject as input representing the serialized yaml file
+func NewFromKubeObject(b *fn.KubeObject) (*NadStruct, error) {
```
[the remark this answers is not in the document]
ganchandrasekaran: At the initialise time, we dont have `CniType` info..  We need to get the interface object and perform validation to confirm the `CniType`. So it is better to skip this param..

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+	p, err := kubeobject.NewFromKubeObject[*nadv1.NetworkAttachmentDefinition](b)
+	if err != nil {
+		return nil, err
+	}
+	return &NadStruct{K: *p}, nil
+}
+
+// NewFromYAML creates a new parser interface
+// It expects a raw byte slice as input representing the serialized yaml file
+func NewFromYAML(b []byte) (*NadStruct, error) {
```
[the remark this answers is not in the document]
ganchandrasekaran: same..

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+	p, err := kubeobject.NewFromYaml[*nadv1.NetworkAttachmentDefinition](b)
+	if err != nil {
+		return nil, err
+	}
+	return &NadStruct{K: *p}, nil
+}
+
+// NewFromGoStruct creates a new parser interface
+// It expects a go struct representing the interface krm resource
+func NewFromGoStruct(x *nadv1.NetworkAttachmentDefinition) (*NadStruct, error) {
```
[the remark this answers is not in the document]
ganchandrasekaran: same..

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+		return r.K.SetNestedString(string(b), ConfigType...)
+	} else {
+		return fmt.Errorf("unknown nad master interface")
+	}
+}
+
+func (r *NadStruct) SetIpamAddress(ipam []Addresses) error {
+	if ipam != nil {
+		nadConfigStruct := r.GetNadConfig()
+		nadConfigStruct.Plugins[0].Ipam.Addresses = ipam
```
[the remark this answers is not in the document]
ganchandrasekaran: On the GetNadConfig, we already make this check.. Also I will refactor this block.. 

`nadConfigStruct.Plugins == nil || len(nadConfigStruct.Plugins) == 0`

krm-functions/lib/nad/v1/nad.go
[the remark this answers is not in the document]
kispaljr: yes, but during our last meeting we agreed that we will change this everywhere, including in the ipalloc lib

krm-functions/lib/nad/v1/nad.go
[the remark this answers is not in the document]
kispaljr: but, if CniTypeSpec is VLAN, then the GetNadConfig may return an empty Plugins list, isn't it?

krm-functions/lib/nad/v1/nad.go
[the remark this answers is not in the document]
kispaljr: In this case would it make sense to introduce an "Uninitialized" option to the enum, to be able to detect that it hasn't been set, and fail methods that rely on the CniType value in this case?

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+		return nil, err
+	}
+	return &NadStruct{K: *p}, nil
+}
+
+func (r *NadStruct) GetCniSpecType(fields ...string) CniSpecType {
+	return r.C
+}
+
+func (r *NadStruct) GetStringValue(fields ...string) string {
```
kispaljr, line 114: do we need this?

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+	nadConfigStruct := NadConfig{}
+	if err := json.Unmarshal([]byte(r.GetStringValue(ConfigType...)), &nadConfigStruct); err != nil {
+		return ""
+	}
+	return nadConfigStruct.Plugins[0].Type
+}
+
+func (r *NadStruct) GetVlan() int {
+	nadConfigStruct := NadConfig{}
+	if err := json.Unmarshal([]byte(r.GetStringValue(ConfigType...)), &nadConfigStruct); err != nil {
```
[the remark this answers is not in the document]
ganchandrasekaran: GetConfigSpec is used to create new configSpec.. So created another fn to avoid repeting json.Unmarshal() call

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+		return ""
+	}
+	return nadConfigStruct.Plugins[0].Master
+}
+
+func (r *NadStruct) GetIpamAddress() []Addresses {
+	nadConfigStruct := NadConfig{}
+	if err := json.Unmarshal([]byte(r.GetStringValue(ConfigType...)), &nadConfigStruct); err != nil {
```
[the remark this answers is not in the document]
ganchandrasekaran: GetConfigSpec is used to create new configSpec.. So created another fn to avoid repeting json.Unmarshal() call

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,283 @@
...
+	if configSpec == "" {
+		configSpec = "{}"
+	}
+	if err := json.Unmarshal([]byte(configSpec), &nadConfigStruct); err != nil {
+		panic(err)
+	}
+	nadConfigStruct.CniVersion = CniVersion
+	if r.GetCniSpecType() == VlanType {
+		return nadConfigStruct
+	} else if r.GetCniSpecType() == IpamType || r.GetCniSpecType() == Both {
```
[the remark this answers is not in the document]
ganchandrasekaran: Already addressed in latest review

krm-functions/lib/nad/v1/nad.go
ganchandrasekaran: commented and refactored

krm-functions/lib/nad/v1/nad.go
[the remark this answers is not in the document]
ganchandrasekaran: If the CniTypeSpec is only VLAN, then this block will not be called at all..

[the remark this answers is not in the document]
ganchandrasekaran: It would be better to check for completnes.. I will add on all the similar blocks..

krm-functions/lib/nad/v1/nad.go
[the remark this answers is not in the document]
ganchandrasekaran: For this purpose I am using "BothIpamVlan" option in emum as default.. So if specific use-case is not found then the generic CniType value is returned

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,364 @@
...
+		if plugin.Type == TuningType {
+			continue
+		} else {
+			return r.GetExistingNadConfig().Plugins[i].Ipam.Addresses
+		}
+	}
+	return []Addresses{}
+}
+
+func (r *NadStruct) GetNewNadConfig() (NadConfig, error) {
```
gvbalaji, line 180: Suggest chaning the name to CreateNewNadConfig

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,364 @@
...
+		}
+		for i, plugin := range nadConfigStruct.Plugins {
+			if plugin.Type == TuningType {
+				continue
+			} else {
+				nadConfigStruct.Plugins[i].Type = cniType
+			}
+		}
+		return r.SetNadConfig(nadConfigStruct)
+	} else {
```
gvbalaji, line 302: Suggest checking for blank value first and return error and then do rest of processing.

krm-functions/lib/nad/v1/nad.go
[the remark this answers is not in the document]
gvbalaji: I am on the fence. Since this is a library would it make sense to return error and let the caller to decide how to handle it?

gvbalaji (commented):
Thanks @ganchandrasekaran . Added a few comments.

krm-functions/lib/nad/v1/nad.go
ganchandrasekaran: removed

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,364 @@
...
+func (r *NadStruct) GetConfigSpec() string {
+	if r == nil {
+		return ""
+	}
+	return r.GetStringValue(ConfigType...)
+}
+
+func (r *NadStruct) GetExistingNadConfig() NadConfig {
+	nadConfigStruct := NadConfig{}
+	if err := json.Unmarshal([]byte(r.GetStringValue(ConfigType...)), &nadConfigStruct); err != nil {
```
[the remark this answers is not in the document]
ganchandrasekaran: returning proper error message to handle

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,366 @@
...
+	nadConfigStruct := NadConfig{}
+	configSpec := r.GetConfigSpec()
+	if configSpec == "" {
+		configSpec = "{}"
+	}
+	if err := json.Unmarshal([]byte(configSpec), &nadConfigStruct); err != nil {
+		return nadConfigStruct, fmt.Errorf("invalid NAD Config, %s", err)
+	}
+	nadConfigStruct.CniVersion = CniVersion
+	if r.CniSpecType == VlanAllocOnly {
```
efiacor, line 192: Hi @ganchandrasekaran,
Quite new to golang and the project so please bear with my noob comments if they aren't helpful.

Would it make sense to use a switch case here?

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,366 @@
...
+		return nadConfigStruct, nil
+	} else if r.CniSpecType == IpVlanType {
+		if nadConfigStruct.Plugins == nil || len(nadConfigStruct.Plugins) == 0 {
+			nadConfigStruct.Plugins = []PluginCniType{
+				{
+					Capabilities: Capabilities{
```
efiacor, line 198: Not sure on the convention here but there seems to be some inconsistencies in the formatting/indentation.

Capabilities: Capabilities{
						Ips: true,
					}

vs

Capabilities: Capabilities{Ips: true}

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,366 @@
...
+						Addresses: []Addresses{
+							{},
+						},
+					},
+				},
+			}
+		}
+		return nadConfigStruct, nil
+	} else if r.CniSpecType == MacVlanType {
+		if nadConfigStruct.Plugins == nil || len(nadConfigStruct.Plugins) == 0 {
```
efiacor, line 213: Quite a bit of repetition here. Is there scope for some refactoring?

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+	return nadConfigStruct, nil
+}
+
+// SetConfigSpec sets the spec attributes in the kubeobject according the go struct
+func (r *NadStruct) SetConfigSpec(spec *nadv1.NetworkAttachmentDefinitionSpec) error {
+	return r.K.SetNestedString(spec.Config, ConfigType...)
+}
+
+func (r *NadStruct) SetNadConfig(config NadConfig) error {
```
efiacor, line 275: Again, convention but should getters and setters be placed together for readability?
ie 
getA()
setA()
getB()
setB()

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,366 @@
...
+	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	"github.com/nephio-project/nephio/krm-functions/lib/kubeobject"
+)
+
+const (
+	errKubeObjectNotInitialized = "KubeObject not initialized"
+	CniVersion                  = "0.3.1"
+	ModeBridge                  = "bridge"
+	ModeL2                      = "l2"
+	NadType                     = "static"
```
efiacor, line 32: Rename to NadTypeStatic|StaticNadType ?

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+type CniSpecType int64
+
+const (
+	OtherType CniSpecType = iota // 0
+	VlanAllocOnly
+	IpVlanType
+	SriovType
+	MacVlanType
```
efiacor, line 76: Not sure what the test coverage goal for the project is but I am getting just under 70%. 
Still learning the tools so may not be an accurate reflection.

krm-functions/lib/nad/v1/nad.go
ganchandrasekaran: @efiacor Hey, thanks for the review. Yes it would..

krm-functions/lib/nad/v1/nad.go
ganchandrasekaran: Current the same method creates a new Config type and also returns a existing config-type for which such if condition is repeted on every switch case..

krm-functions/lib/nad/v1/nad.go
ganchandrasekaran: Got it over 80%, the rest are simply struct and const definitions..

krm-functions/lib/nad/v1/nad.go
ganchandrasekaran: Aligned the functions similar to other libs..

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,366 @@
...
+					},
+				},
+				{
+					Capabilities: Capabilities{Mac: true},
+					Type:         TuningType,
+				},
+			}
+		}
+		return nadConfigStruct, nil
+	} else if r.CniSpecType == SriovType {
```
henderiw, line 232: is this not better through a switch statement

henderiw:
Focus on R1 content in this PR. A new PR should cover making this more generic.

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,370 @@
...
+				nadConfigStruct.Plugins[i].Type = cniType
+			}
+		}
+		return r.setNadConfig(nadConfigStruct)
+	}
+}
+
+func (r *NadStruct) SetVlan(vlanType int) error {
+	if vlanType != 0 {
+		nadConfigStruct, err := r.getNadConfig()
```
henderiw, line 315: better to return err and fail fast

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,370 @@
...
+		}
+		nadConfigStruct.Vlan = vlanType
+		return r.setNadConfig(nadConfigStruct)
+	} else {
+		return fmt.Errorf("unknown vlanType")
+	}
+}
+
+func (r *NadStruct) SetNadMaster(nadMaster string) error {
+	if nadMaster != "" {
```
henderiw, line 327: do the error first and fail fast

krm-functions/lib/nad/v1/nad.go
```diff
@@ -0,0 +1,370 @@
...
+			}
+		}
+		return r.setNadConfig(nadConfigStruct)
+	} else {
+		return fmt.Errorf("unknown nad master interface")
+	}
+}
+
+func (r *NadStruct) SetIpamAddress(ipam []Addresses) error {
+	if ipam != nil {
```
henderiw, line 346: same fail fast

krm-functions/lib/nad/v1/nad.go
ganchandrasekaran: changes here: https://github.com/nokia/nephio-nephio/pull/27/files

aakashchan (approved):

PR #68, opened 2023-03-30
added clustercontext lib

henderiw:
cluster context lib helps the arm functions on operations of the clustercontext crd


krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
```diff
@@ -0,0 +1,83 @@
...
+	// The structure of the generated document will reflect the structure of the value itself.
+	Marshal() ([]byte, error)
+	// ParseKubeObject returns a fn sdk KubeObject; if something failed an error
+	// is returned
+	ParseKubeObject() (*fn.KubeObject, error)
+}
+
+// NewMutator creates a new mutator for the ClusterContext
+// It expects a raw byte slice as input representing the serialized yaml file
+func NewMutator(b string) ClusterContext {
```
gvbalaji, line 41: Similar to other PR where I have commented on this:
1. Please consider changing the name of the method NewMutator to NewClusterContext or something like that.
2. Construct the full object by deserializing the byte stream in this method so that we have the object completely initialized.

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
```diff
@@ -0,0 +1,83 @@
...
+		raw: []byte(b),
+	}
+}
+
+type clusterContext struct {
+	raw     []byte
+	cluster *infrav1alpha1.ClusterContext
+}
+
+func (r *clusterContext) UnMarshal() (*infrav1alpha1.ClusterContext, error) {
```
gvbalaji, line 52: If we initialize the object in the New method as said above this method will become a getter method.

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := yaml.Unmarshal(r.raw, c); err != nil {
+		return nil, err
+	}
+	r.cluster = c
+	return c, nil
+}
+
+// Marshal serializes the value provided into a YAML document based on "sigs.k8s.io/yaml".
+// The structure of the generated document will reflect the structure of the value itself.
+func (r *clusterContext) Marshal() ([]byte, error) {
```
gvbalaji, line 63: If we initialize the object in the New method as said above this method will become a getter method.

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+	r.raw = b
+	return b, err
+}
+
+// ParseKubeObject returns a fn sdk KubeObject; if something failed an error
+// is returned
+func (r *clusterContext) ParseKubeObject() (*fn.KubeObject, error) {
+	b, err := r.Marshal()
```
gvbalaji, line 78: Why are we calling Marshall method from here when the object is already initialized with byte stream?

krm-functions/lib/clustercontext/v1alpha1/clustercontext_test.go
```diff
@@ -0,0 +1,17 @@
+/*
```
gvbalaji, line 1: I guess you will add tests.

gvbalaji (changes requested):
Thanks for the PR. Great learning for us. Added a few comments.

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
henderiw: fixed and aligned with the other functions

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
henderiw: fixed

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
henderiw: this is hidden now

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
henderiw: I no longer keep the byte slice as it is used only for initialisation. Afterwards we expect people to interact with the gostruct. The ParseKubeObject is typically used to convert the latest go struct to a fn.KubeObject struct.

krm-functions/lib/clustercontext/v1alpha1/clustercontext_test.go
henderiw: done

makefile
```diff
@@ -0,0 +1,11 @@
+.PHONY: all
```
gvbalaji, line 1: May need to sync the makefile.

gvbalaji (approved):
Looks good. Thanks @henderiw .

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/GoogleContainerTools/kpt v1.0.0-beta.29.0.20230327202912-01513604feaa
 	github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20230302070146-e8e9cb3c3ae2
 	github.com/google/go-cmp v0.5.9
+	github.com/nephio-project/nephio-controller-poc v0.0.2
```
johnbelamaric, line 9: If we are grabbing a type from here, we should move it to nephio-project/api instead.

go.sum
johnbelamaric, line 1: As noted in the other PR, let's avoid version updates (pulling in necessary things of course is another story)

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
```diff
@@ -0,0 +1,77 @@
+/*
+ Copyright 2023 Nephio.
```
johnbelamaric, line 2: "Nephio Project Authors"

makefile
johnbelamaric, line 1: remove this, shouldn't be needed any more (a rebase will be needed though)

krm-functions/lib/clustercontext/v1alpha1/clustercontext.go
```diff
@@ -0,0 +1,77 @@
...
+	b, err := yaml.Marshal(r.cluster)
+	if err != nil {
+		return nil, err
+	}
+	return b, err
+}
+
+// ParseKubeObject returns a fn sdk KubeObject; if something failed an error
+// is returned
+func (r *clusterContext) ParseKubeObject() (*fn.KubeObject, error) {
```
johnbelamaric, line 66: See comments in the Interface version - I think: 1) we need to do this differently to preserve YAML comments; 2) we can probably do the heavy lifting with one bit of code, and then each specific library per type just does type casting...I think.

johnbelamaric (changes requested):

PR #69, opened 2023-03-30
interface-fn 

henderiw:
interface-fn is a function in Nephio R1 that operates on the interface requirement CR. It performs CRUD operation in the package on the NAD, VLANALlocation and IP Allocation resources.


gvbalaji:
@henderiw thanks. I understand this will change once the function runtime library and it's client intarface are finalized and merged. If my understanding is correct, can you please convert this into a draft PR?

kispaljr:
@henderiw I believe you failing gosec test can be solved by go workspaces (as You suggested on Slack :) )
This is how I did it with dnn-fn: https://github.com/nephio-project/nephio/pull/70, notice the go.work file in root of the repo.

krm-functions/interface-fn/mutator.go
```diff
@@ -0,0 +1,269 @@
...
+		rl.Results.ErrorE(err)
+		return false, nil
+	}
+	return myFn.sdk.Run()
+}
+
+// ClusterContextCallbackFn provides a callback for the cluster context
+// resources in the resourceList
+func (r *itfceFn) ClusterContextCallbackFn(o *fn.KubeObject) error {
+	clusterKOE, err := kubeobject.NewFromKubeObject[*infrav1alpha1.ClusterContext](o)
```
kispaljr, line 75: This comment is just a reminder for a potential need for change in the future:
As I wrote on Slack we should decide whether to use pointer or struct type parameters of KubeObjectExt.
If the decision will be pointers, than no comment here, otherwise the code here and other places should be aligned to the decision.

krm-functions/interface-fn/mutator.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+	}
+	vlanallocs := objs.Where(fn.IsGroupVersionKind(vlanv1alpha1.VLANAllocationGroupVersionKind))
+	for _, vlanalloc := range vlanallocs {
+		if vlanalloc.GetName() == forObj.GetName() {
+			alloc, err := vlanlibv1alpha1.NewFromKubeObject(vlanalloc)
+			if err != nil {
+				return nil, err
+			}
+			//alloc, err := ko.NewFromKubeObject[*vlanv1alpha1.VLANAllocation](vlanalloc)
```
kispaljr, line 203: this comment is not needed any more

kispaljr (approved):
As I only have cosmetic comments, This looks good to me even in it current state

krm-functions/interface-fn/fn/function.go
```diff
@@ -0,0 +1,270 @@
...
+			//}
+			allocGoStruct, err := alloc.GetGoStruct()
+			if err != nil {
+				return nil, err
+			}
+			itfce.Status.VLANAllocationStatus = &allocGoStruct.Status
+		}
+	}
+	// set the status
+	err = itfceKOE.SetStatus(itfce.Status)
```
kispaljr, line 216: With the new API this should be:
`itfceKOE.SetStatus(itfce)`

kispaljr (changes requested):
Nested go.mod files bites us again:
presubmit-nephio-go-test  doesn't run any tests inside `krm-functions/interface-fn` at all.
Actually since the new (not backward compatible) KubeObjectExt API was merged into this branch, this code doesn't even compile, and none of our presubmit tests detected this :(
I've left comments for the quick changes required for the code to compile, otherwise this LGTM

gvbalaji:
Thanks  @henderiw . A rebase is needed for this PR.

PR #70, opened 2023-03-30
dnn-fn

henderiw:
DNN-fn operates on DNN resource and can request IP Allocations.


krm-functions/dnn-fn/main.go
```diff
@@ -0,0 +1,183 @@
...
+			Owns: map[corev1.ObjectReference]kptcondsdk.ResourceKind{
+				{
+					APIVersion: ipamv1alpha1.GroupVersion.Identifier(),
+					Kind:       ipamv1alpha1.IPAllocationKind,
+				}: kptcondsdk.ChildRemote,
+			},
+			Watch: map[corev1.ObjectReference]kptcondsdk.WatchCallbackFn{
+				{
+					APIVersion: infrav1alpha1.GroupVersion.Identifier(),
+					Kind:       "ClusterContext",
```
ganchandrasekaran, line 64: Suggestion: instead of embeded "ClusterContext" string you can use below:

Kind:       reflect.TypeOf(infrav1alpha1.ClusterContext{}).Name(),

krm-functions/dnn-fn/main.go
```diff
@@ -0,0 +1,183 @@
...
+	return true, nil
+}
+
+func (f *DnnFn) ClusterContextCallbackFn(o *fn.KubeObject) error {
+	var cluster infrav1alpha1.ClusterContext
+	err := o.As(&cluster)
+	if err != nil {
+		return err
+	}
+	f.site = *cluster.Spec.SiteCode
```
ganchandrasekaran, line 85: Here if the siteCode field is not available on clustercontext yaml then following occurs:
panic: runtime error: invalid memory address or nil pointer dereference

it would be better to handle it here..

henderiw (commented):
why are we using the runtime scheme if we can access the schema direct. Avoiding it removes some code

krm-functions/dnn-fn/main.go
```diff
@@ -0,0 +1,194 @@
...
+	ipam_common "github.com/nokia/k8s-ipam/apis/alloc/common/v1alpha1"
+	ipamv1alpha1 "github.com/nokia/k8s-ipam/apis/alloc/ipam/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// var _ fn.Runner = &DnnFn{}
+
+var theScheme *runtime.Scheme = runtime.NewScheme()
```
henderiw, line 38: why do we add this here? we can access the schema direct w/o going to runtime scheme. avoids some code

krm-functions/dnn-fn/main.go
```diff
@@ -0,0 +1,194 @@
...
+
+// TODO: this should come from infrav1alpha1
+type DnnPoolStatus struct {
+	Name string                          `yaml:"name,omitempty" json:"name,omitempty"`
+	Ip   ipamv1alpha1.IPAllocationStatus `yaml:"ip,omitempty" json:"ip,omitempty"`
+}
+
+// GetKindOrPanic returns with the Kind of a Kubernetes API resource type `T`.
+// Panics if `T` is not registered in the `theScheme`
+func GetKindOrPanic[T any, PT interface {
```
henderiw, line 59: can be avoided if we don't go to the scheme

krm-functions/dnn-fn/main.go
```diff
@@ -0,0 +1,194 @@
...
+		err := o.As(&ipalloc)
+		if err != nil {
+			return nil, err
+		}
+
+		poolName, found := strings.CutPrefix(ipalloc.Name, dnnObj.GetName()+"-")
+		if found {
+			pools = append(pools, DnnPoolStatus{Name: poolName, Ip: ipalloc.Status})
+		} else {
+			f.rl.Results.Warningf("found an owned IPAllocation with a suspicious name: %v", ipalloc.Name)
```
henderiw, line 178: is it not better to log rather than adding to resourceList? I would like to avoid adding to resourceList.

henderiw:
Need to discuss the test strategy if adding all these files is a good approach or not

kispaljr:
> Need to discuss the test strategy if adding all these files is a good approach or not

Based on our discussion yesterday I've created a PR for moving my test helper functions into lib/: https://github.com/nephio-project/nephio/pull/134

henderiw:
@kispaljr can you look at the build issues ?

gvbalaji:
Thanks @henderiw . A rebase needed for this PR.

kispaljr:
Tests will pass only after https://github.com/nephio-project/nephio/pull/170 is merged

krm-functions/dnn-fn/lib.go
```diff
@@ -0,0 +1,100 @@
...
+	infrav1alpha1 "github.com/nephio-project/api/infra/v1alpha1"
+	nephioreqv1alpha1 "github.com/nephio-project/api/nf_requirements/v1alpha1"
+	ipamv1alpha1 "github.com/nokia/k8s-ipam/apis/alloc/ipam/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var theScheme *runtime.Scheme = runtime.NewScheme()
+
+func init() {
```
henderiw, line 33: do you still need the schema stuff?

krm-functions/dnn-fn/testdata/golden/garbage_collection/ipallocations.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: internet-pool1
+  annotations:
+    specializer.nephio.org/owner: req.nephio.org/v1alpha1.DataNetwork.internet
+spec:
+  kind: pool
+  selector:
+    matchLabels:
+      nephio.org/site: edge1
+  networkInstance: {}
+  prefixLength: 8
```
gvbalaji, line 13: newline

krm-functions/dnn-fn/lib.go
henderiw: Here is an example how I did it in the interface-fn. You basically know the types based on GVK as it is 1:1 mapping

ipallocs := objs.Where(fn.IsGroupVersionKind(ipamv1alpha1.IPAllocationGroupVersionKind))
	for _, ipalloc := range ipallocs {
		if ipalloc.GetName() == forObj.GetName() {
			alloc, err := kubeobject.NewFromKubeObject[ipamv1alpha1.IPAllocation](ipalloc)
			if err != nil {
				return nil, err
			}
			allocGoStruct, err := alloc.GetGoStruct()
			if err != nil {
				return nil, err
			}
			itfce.Status.IPAllocationStatus = &allocGoStruct.Status
		}
	}

krm-functions/dnn-fn/main.go
```diff
@@ -0,0 +1,163 @@
...
+	rl             *fn.ResourceList
+}
+
+// Run is the entry point of the KRM function (called by the upstream fn SDK)
+func Run(rl *fn.ResourceList) (bool, error) {
+	var err error
+	myFn := DnnFn{rl: rl}
+
+	// get ClusterContext
+	myFn.ClusterContext, err = GetSingleton[infrav1alpha1.ClusterContext](rl.Items)
```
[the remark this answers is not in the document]
kispaljr: This was meant to be a temporary solution until the singleton functionality we talked about will be part of the SDK. Is it already available? I will update this to the SDK if it is available.

kispaljr:
GitHub working strange to me, somehow I have trouble replying to this, so I will just copy-paste:

Here is an example how I did it in the interface-fn. You basically know the types based on GVK as it is 1:1 mapping

```
ipallocs := objs.Where(fn.IsGroupVersionKind(ipamv1alpha1.IPAllocationGroupVersionKind))
for _, ipalloc := range ipallocs {
if ipalloc.GetName() == forObj.GetName() {
alloc, err := kubeobject.NewFromKubeObject[ipamv1alpha1.IPAllocation](https://github.com/nephio-project/nephio/pull/ipalloc)
if err != nil {
return nil, err
}
allocGoStruct, err := alloc.GetGoStruct()
if err != nil {
return nil, err
}
itfce.Status.IPAllocationStatus = &allocGoStruct.Status
}
}
```

Yes, exactly. `FilterByType[IPAllocation]` implements the exact same functionality as the first 12 lines of your code. Since this code is copy-pasted into all of our KRM functions, I thought it is prime candidate to factor it out to a utility function.
Does this cause a problem?

krm-functions/dnn-fn/main.go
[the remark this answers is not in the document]
kispaljr: I looked, and didn't find, so I am reverting back to the old Watch callback-based version

krm-functions/dnn-fn/lib.go
ganchandrasekaran: Btw with latest lib, the above doesn't work instead it has to be: 
`objs.Where(fn.IsGroupVersionKind(schema.GroupVersionKind(ipamv1alpha1.IPAllocationGroupVersionKind)))`

krm-functions/dnn-fn/main.go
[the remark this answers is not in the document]
henderiw: we can add singleton to the sdk. it would be better to have it there

krm-functions/dnn-fn/README.md
```diff
@@ -0,0 +1,28 @@
...
+`dnn-fn` is a KRM function with two main purposes:
+- for each IP `pool` specified in each `DataNetwork` resource in the kpt package, `dnn-fn` will create a corresponding `IPAllocation`. 
+- for each `IPAllocation` that was created by a previous call of `dnn-fn` it will copy the result of the IP allocation to the `status` of the original `DataNetwork` resource.
+
+<!--mdtogo-->
+
+
+<!--mdtogo:Long-->
+
+## Usage
```
ganchandrasekaran, line 16: Just a suggestion, for the sake of alignment what do you think about changing the content of `usage` sub-section? 

https://github.com/nephio-project/nephio/tree/main/krm-functions/interface-fn#usage
https://github.com/nephio-project/nephio/blob/main/krm-functions/ipam-fn/README.md
https://github.com/nephio-project/nephio/blob/main/krm-functions/nad-fn/README.md#usage

This section can be names as Implementation details, perhaps?

PR #72, opened 2023-03-30
nf-deploy-fn

henderiw:
nf-deploy fn generates the NF Deployment for the respective NFTYpe


gvbalaji:
Looks like this PR is not in use. If so can you please close this. Thanks @henderiw .

PR #73, opened 2023-03-30
Add the initial version of the Interface management KRM function design

kispaljr:


krm-functions/interface-fn/design.md
```diff
@@ -0,0 +1,157 @@
...
+            masterInterface: eth1
+        siteCode: edge1
+        region: us-central1
+
+the following list of CRs will be added to package:
+
+    apiVersion: ipam.nephio.org/v1alpha1
+    kind: IPAllocation
+    metadata:
+      name: n6-ip-abcd
```
ganchandrasekaran, line 51: Can we also include the following as part of metadata:
As compliant with:- https://docs.google.com/document/d/1wAgPO0IDyN-UqpB3twv3e-9mumck-5o3j8f74FLkKPI/ 
```
  labels:
    nephio.org/interface: n3
```

If not, is there a field that has only interface details?

krm-functions/interface-fn/design.md
```diff
@@ -0,0 +1,157 @@
...
+              status: false
+            - type: ipam-nephio-org-v1alpha1-vlanallocation-n6-vlan-defg
+              status: false
+            - type: req-nephio-org-v1alpha1-interface-n6-nad-generated
+              status: false
+
+
+Implementation details
+----------------------
+
```
gvbalaji, line 110: Once the function runtime library client API is decided, it would be good to mention how that would be used here as well.

gvbalaji (commented):
Thanks a lot for the write up @kispaljr . Added a comment ( mostly for future)

PR #84, opened 2023-04-01
added parser interface

henderiw:
This is a generic parser interface that would be used by all libraries that operation as krm functions


johnbelamaric:
I was thinking something like this would be in the nephio/api repo

henderiw:
this is part of the library and we created a lib under krm-functions as the code is more close to the functions versus the api.
In the api repo we have added the getters, but the this is something to consume a library in a fn. So we believe this is a better fit in the function library.

So we beleive something like this make sense:
nephio/krm-functions/lib where a set of helpers libraries reside.

gvbalaji (approved):
Thanks. /lgtm

johnbelamaric:
Can you rebase?

krm-functions/lib/parser/parser.go
```diff
@@ -0,0 +1,24 @@
+/*
+Copyright 2023 Nephio.
```
johnbelamaric, line 2: Let's standardize on "The Nephio Authors"

Then I can merge this

PR #93, opened 2023-04-07
Fix the gosec prow configuration

johnbelamaric:


.prow.yaml
```diff
@@ -16,9 +16,9 @@ presubmits:
       containers:
       - image: securego/gosec:2.15.0
         command:
-        - make
-        args:
         - gosec
+        args:
```
vjayaramrh, line 20: Nice, thanks for the fix

vjayaramrh (approved):

PR #111, opened 2023-04-11
Simplified prow job definitions, changed to use home-brewed gotests image increased lint timeout

radoslawc:


gvbalaji (approved):
approve

.prow.yaml
```diff
@@ -1,33 +1,31 @@
 presubmits:
   - name: presubmit-nephio-go-test
     decorate: true
-    run_if_changed: "(\\.go|^Makefile)$"
+    run_if_changed: "^.*.go$"
     spec:
       containers:
-      - image: golang:1.20.2
+      - image: nephio/gotests:2
```
vjayaramrh, line 7: curious as to what the gotests image adds that is not already in the official golang:1.20.2 image? Thanks

PR #118, opened 2023-04-13
validation fn consumed from interface api

henderiw:
Now that https://github.com/nephio-project/api/pull/10 is merged we use the validation from this library.
moved the nil check back into the main code.


krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -216,6 +234,7 @@ func (r *itfce) getStringValue(fields ...string) string {
 
 // setNestedField is a generic utility function that sets a string on
 // a string slice representing the path in the yaml doc
+/*
```
johnbelamaric, line 237: if we don't need these functions anymore, we should just remove them.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -106,60 +106,75 @@ func (r *itfce) GetGoStruct() (*nephioreqv1alpha1.Interface, error) {
...
 		return nil
 	}
+
+	// validate the spec
+	if err := nephioreqv1alpha1.ValidateInterfaceSpec(spec); err != nil {
+		return err
+	}
+
+	// set or delete the values in the spec based on the information
+	fmt.Println("spec.AttachmentType", spec.AttachmentType)
```
johnbelamaric, line 177: debugging cruft?

johnbelamaric (changes requested):

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -106,60 +106,75 @@ func (r *itfce) GetGoStruct() (*nephioreqv1alpha1.Interface, error) {
 // GetAttachmentType returns the attachmentType from the spec
 // if an error occurs or the attribute is not present an empty string is returned
 func (r *itfce) GetAttachmentType() string {
+	if r.o == nil {
```
gvbalaji, line 109: I think this was using a common function. This block is very common for all getters
{
if r.o == nil {
		return ""
	}
	return r.getStringValue(foo...)
}

My suggestion is to handle the nil condition in common function itself.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -106,60 +106,75 @@ func (r *itfce) GetGoStruct() (*nephioreqv1alpha1.Interface, error) {
...
+		return ""
+	}
 	return r.getStringValue(attachmentType...)
 }
 
 // GetCNIType returns the cniType from the spec
 // if an error occurs or the attribute is not present an empty string is returned
 func (r *itfce) GetCNIType() string {
+	if r.o == nil {
```
gvbalaji, line 118: Same as above, handle the nil in the common function.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -106,60 +106,75 @@ func (r *itfce) GetGoStruct() (*nephioreqv1alpha1.Interface, error) {
...
+		return ""
+	}
 	return r.getStringValue(cniType...)
 }
 
 // GetNetworkInstanceName returns the name of the networkInstance from the spec
 // if an error occurs or the attribute is not present an empty string is returned
 func (r *itfce) GetNetworkInstanceName() string {
+	if r.o == nil {
```
gvbalaji, line 127: Same as above, handle the nil in the common function.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -106,60 +106,75 @@ func (r *itfce) GetGoStruct() (*nephioreqv1alpha1.Interface, error) {
...
-	default:
+	if r.o == nil {
+		return fmt.Errorf(errKubeObjectNotInitialized)
+	}
+	if !nephioreqv1alpha1.IsAttachmentTypeSupported(s) {
 		return fmt.Errorf("unknown attachmentType")
 	}
 
-	return r.setNestedField(s, attachmentType...)
+	return r.o.SetNestedField(s, attachmentType...)
```
gvbalaji, line 142: My suggestion is to use the common function for Sets as before as well.
1. Validate which is specific to the method.
2. Call the common setter that will do the set.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -106,60 +106,75 @@ func (r *itfce) GetGoStruct() (*nephioreqv1alpha1.Interface, error) {
...
 }
 
 // SetCNIType sets the cniType in the spec
 func (r *itfce) SetCNIType(s string) error {
```
gvbalaji, line 146: Same as other setter. Call the common function.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -106,60 +106,75 @@ func (r *itfce) GetGoStruct() (*nephioreqv1alpha1.Interface, error) {
...
 		return fmt.Errorf("unknown cniType")
 	}
 
-	return r.setNestedField(s, cniType...)
+	return r.o.SetNestedField(s, cniType...)
 }
 
 // SetNetworkInstanceName sets the name of the networkInstance in the spec
 func (r *itfce) SetNetworkInstanceName(s string) error {
-	return r.setNestedField(s, networkInstanceName...)
```
gvbalaji, line 155: Same as other sets.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -182,28 +197,31 @@ func (r *itfce) SetSpec(spec *nephioreqv1alpha1.InterfaceSpec) error {
...
-	} else {
-		return fmt.Errorf("networkInstance is required")
 	}
 	return nil
 }
 
 // DeleteAttachmentType deletes the attachmentType from the spec
 func (r *itfce) DeleteAttachmentType() error {
-	return r.deleteNestedField(attachmentType...)
+	if r.o == nil {
```
gvbalaji, line 206: Same as before. I suggest using a common delete.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -182,28 +197,31 @@ func (r *itfce) SetSpec(spec *nephioreqv1alpha1.InterfaceSpec) error {
...
+		return fmt.Errorf(errKubeObjectNotInitialized)
+	}
+	_, err := r.o.RemoveNestedField(cniType...)
+	return err
 }
 
 // DeleteAttachmentType deletes the attachmentType from the spec
 func (r *itfce) DeleteCNIType() error {
-	return r.deleteNestedField(cniType...)
+	if r.o == nil {
```
gvbalaji, line 215: Use common delete

krm-functions/lib/interface/v1alpha1/interface.go
gvbalaji: My suggestion is to use them as this is common boiler plate code.

gvbalaji (changes requested):
Thanks @henderiw . Added a few comments. Basically suggesting to use the common methods as you had done before.

krm-functions/lib/interface/v1alpha1/interface.go
vjayaramrh: FYI There was code being processed earlier without checking for the nil condition and had mentioned this to Wim. I made the point that if check for nil fails, no need to do further processing. @gvbalaji Maybe, it would be a good idea for you, Wim and I to discuss this quick to avoid further confusion? What do you say?

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: removed

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -182,28 +176,23 @@ func (r *itfce) SetSpec(spec *nephioreqv1alpha1.InterfaceSpec) error {
...
-	} else {
-		return fmt.Errorf("networkInstance is required")
 	}
 	return nil
 }
 
 // DeleteAttachmentType deletes the attachmentType from the spec
 func (r *itfce) DeleteAttachmentType() error {
-	return r.deleteNestedField(attachmentType...)
+	return r.deleteNestedField(cniType...)
```
gvbalaji, line 185: Looks like a typo here. "attachmentType" needs to be passed in.

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -182,28 +176,23 @@ func (r *itfce) SetSpec(spec *nephioreqv1alpha1.InterfaceSpec) error {
...
 }
 
-// DeleteAttachmentType deletes the attachmentType from the spec
+// DeleteCNIType deletes the cniType from the spec
 func (r *itfce) DeleteCNIType() error {
-	return r.deleteNestedField(cniType...)
+	return r.deleteNestedField(attachmentType...)
```
gvbalaji, line 190: Looks like a typo here. "cniType" needs to be passed in.

gvbalaji (changes requested):

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: fixed + added test to cover this

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: fixed + added a test to cover this

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: aligned

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: aligned

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: aligned

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: agreed

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: With the current flow we valid input first and do nil check after if we apply to the object. At this stage the code is most optimal.

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: fixed

PR #121, opened 2023-04-16
added new generic parser interface

henderiw:
this is the generic parser interface which offloads a set of method implementations from the libraries


gvbalaji (changes requested):
Thanks @henderiw . My initial thinking is the following. I have updated this after seeing how this is consumed in the interface lib. Here are my suggestions.
1. I think it may be simpler to have only helper functions in this base library. 

type Entity struct {
	Obj *fn.KubeObject
}

- NewFromKubeObject(*fn.KubeObject)  (Entity, error)
- NewFromYaml(b []byte) (Entity, error)
- NewFromGoStruct(x any) (Entity, error) 
- GostructFromKubeObject[T any](*fn.KubeObject) (T, err)



This should cover the creation of objects for all entities ( ip, interface etc) and base methods. Then we do not need to implement in the individual libraries.


3. Let the callers directly call get/set/delete methods on the kubeobject. I am saying this because we are not providing any additional processing (for e.g like we did with GetInterface... methods to hide the YAML path from the user). 
4. Then in individual libraries only define getters/setters that will do further help like in the interface library  GetAttachmentType(), GetNetworkInstanceName() etc. Implement these methods directly on the Entity structure defined above ( in this base library). This will avoid lot of code in each of the libraries.

in interface library

type InterfaceEntity interface {
GetAttachmentType()
GetNetworkInstanceName()
SetNetworkInstanceName()
.
.
.
}


func (r *Entity) GetAttachmentType() string
func (r *Entity) GetNetworkInstanceName() string

Users will first create the entity  and then call the methods based on function they are implementing.

krm-functions/lib/parser/parser.go
```diff
@@ -17,11 +17,218 @@ limitations under the License.
...
+	SetNestedMap(s map[string]string, fields ...string) error
+	// DeleteNestedField is a generic utility function that deletes
+	// a string slice representing the path from the yaml doc
+	DeleteNestedField(fields ...string) error
+}
+
+// NewFromKubeObject creates a new parser interface
+// It expects a *fn.KubeObject as input representing the serialized yaml file
+func NewFromKubeObject[T1 any](o *fn.KubeObject) Parser[T1] {
+	return &obj[T1]{
```
gvbalaji, line 68: May be a good idea to check  input "o" for nil and return error. Since this is a base method will be useful for all callers.

kispaljr:
I agree with @gvbalaji to have only helper functions in the base library. 
To help the argument, this: https://github.com/nokia/nephio-nephio/pull/19 alternative (but functionally equivalent) implementation of the same  `Parser` interface shows, that most of the Parser methods are just aliases of KubeObject methods.

The same implementation also shows that the 4 functions @gvbalaji mentioned, are also mostly aliases to existing upstream SDK functions. So in his proposal we only need conversion functions between Entity and KubeObject. Or, we can just use `type Entity fn.KubeObject` instead of the `type Entity struct {...}`  to make it trivial.

krm-functions/lib/kubeobject/kubeobject.go
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+package parser
+
+import (
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	"sigs.k8s.io/yaml"
+)
+
+type KubeObject[T1 any] struct {
```
gvbalaji, line 24: Should we change the name to KubeObjectExt ( meaning extension). Naming is tough:). Just wanted to distinguish from the base.

gvbalaji (commented):
Thanks @henderiw . In line with our discussion. Added a minor comment about naming.

krm-functions/lib/kubeobject/kubeobject.go
```diff
@@ -0,0 +1,60 @@
...
+	var x T1
+	if err := yaml.Unmarshal([]byte(r.String()), &x); err != nil {
+		return x, err
+	}
+	return x, nil
+}
+
+// NewFromKubeObject returns a KubeObject struct
+// It expects a fn.KubeObject as input representing the serialized yaml file
+func NewFromKubeObject[T1 any](o fn.KubeObject) *KubeObject[T1] {
```
gvbalaji, line 38: Probably we can check for nil input and return err. Makes it consistent with the rest .

krm-functions/lib/kubeobject/kubeobject.go
henderiw: done

krm-functions/lib/kubeobject/kubeobject.go
henderiw: done

gvbalaji (approved):
/lgtm

PR #122, opened 2023-04-16
Iflib newKO

henderiw:
aligned the interface with the new parser implementation
added setters using the real types as this eases validation.


kispaljr:
My comment for the IPAllocation resource lib, also applies here: 
https://github.com/nephio-project/nephio/pull/123#pullrequestreview-1387931930

gvbalaji:
Please see my comments on https://github.com/nephio-project/nephio/pull/121 which is base for others. I beleive if we take that path implementation of these individual libraries will be simpler.

kispaljr:
Just wanted to point out that this PR contains files from #121 (that this PR depends on). This may be fine, but it can also trouble if this is merged after #121. 
I am not sure what our policy is in this scenario.

gvbalaji:
Thanks @kispaljr . Yes a rebase with main needed . In addition as discussed this PR will change in many ways and be simpler as we discussed. This will only have methods directly implemented on `KubeObjectEx` . Those methods will be only a few handful. All the kubeobject create methods are handled in base parser as well.

gvbalaji (commented):
vThanks @henderiw . I was hoping we will avoid all "Newfromxxx "methods in this file. Basically instead of creating a new structure, add the interface specific methods directly on KubeObjectEx  created in the base parser( few get/set/deletes). Then the usage pattern would be callers first create the object using the base parser and call respective methods based on which function they are implementing. That way these libraries will be simpler. What do you think?

johnbelamaric:
It's more idiomatic in Go to avoid field setters and getters. So IMO, methods to create from YAML or Go struct, and to convert to Go struct, and to set via Go struct are all we need, as described here: https://github.com/nephio-project/nephio/pull/123#pullrequestreview-1398837764

henderiw:
Aligned the interface lib with the new setSpec library in kubeobject

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -17,224 +17,41 @@
...
-	SetNetworkInstanceName(s string) error
-	// SetSpec sets the spec attributes in the kubeobject according the go struct
-	SetSpec(*nephioreqv1alpha1.InterfaceSpec) error
-	// DeleteAttachmentType deletes the attachmentType from the spec
-	DeleteAttachmentType() error
-	// DeleteAttachmentType deletes the attachmentType from the spec
-	DeleteCNIType() error
+// NewFromKubeObject creates a new parser interface
+// It expects a *fn.KubeObject as input representing the serialized yaml file
+func NewFromKubeObject(o *fn.KubeObject) (*Interface, error) {
```
gvbalaji, line 31: In the same spirit as above comment, show we just return the base KubeObjectExt from here?
In fact I am also thinking can the function devs directly invoke kubeobject.NewFromKubeObject[*vlanv1alpha1. Interface]?
That could mean do we really need this library for indirection?

gvbalaji (commented):
Thanks @henderiw . I am on the fence if we really need this with all the changes we made to the base KubeObjectExt. Added a couple of comments.
I think the tests for SetSpec and SetStatus are useful. Even  if we decide to do away from this library, I suggest we keep the tests by moving them to the base KubeObjectExt .

krm-functions/lib/interface/v1alpha1/interface.go
```diff
@@ -17,224 +17,41 @@
...
-	errKubeObjectNotInitialized = "KubeObject not initialized"
+	"github.com/nephio-project/nephio/krm-functions/lib/kubeobject"
 )
 
-var (
-	attachmentType      = []string{"spec", "attachmentType"}
-	cniType             = []string{"spec", "cniType"}
-	networkInstanceName = []string{"spec", "networkInstance", "name"}
-)
+type Interface struct {
```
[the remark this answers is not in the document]
henderiw: The value is you get explicit type checking for the object and we need this in several libraries. This is the reason to do this. otherwise people need to implement this in each function independently again and again.

krm-functions/lib/interface/v1alpha1/interface.go
henderiw: The value is you get explicit type checking for the object and we need this in several libraries. This is the reason to do this. otherwise people need to implement this in each function independently again and again.

henderiw: what I did not do so far is have the explicit SeptSpec and SetStatus with their specific type, so this would add even more type awareness.

krm-functions/lib/interface/v1alpha1/interface.go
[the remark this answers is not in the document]
henderiw: what I did not do so far is have the explicit SeptSpec and SetStatus with their specific type, so this would add even more type awareness.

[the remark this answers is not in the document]
henderiw: the way to look at this is a the specific KubeObj extensions for the InterfaceReq CRD

krm-functions/lib/interface/v1alpha1/interface.go
gvbalaji: thanks @henderiw . Approved

kispaljr:
If I understand correctly, that the main reason to have this wrapper at all, is the lack of static type checking in the interface{} type parameter of SetSpec and SetStatus. I agree that this is a valid concern, however I suggest to use the power of Go generics to fix this, as opposed to implementing it one-by-one for each type (avoiding this is the main reason for having generics).
Here s my proposal to add static type checking: https://github.com/nephio-project/nephio/pull/141

PR #125, opened 2023-04-18
conditional dance sdk

henderiw:
this is the code for the sdk that offer the conditional dance.


krm-functions/lib/condkptsdk/sdk.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (r *sdk) Run() {
+	if r.rl.Items.Len() == 0 {
+		r.rl.Results = append(r.rl.Results, fn.ErrorResult(fmt.Errorf("no resources present in the resourcelist")))
+		return
+	}
+	// get the kptfile first as we need it in various places
+	// we assume the kpt file is always resource idx 0 in the resourcelist
+	var err error
+	r.kptf, err = kptfilelibv1.New(r.rl.Items.GetRootKptfile().String())
```
kispaljr, line 90: GetRootKptfile() retruns with nil, if there is no Kptfile resource in the package. This line panics with nil pointer dereference in this case, that may be replaced by a more user friendly error message.

henderiw: thx indeed this was not the right way to handle the error.
Fixed

krm-functions/lib/condkptsdk/design.md
```diff
@@ -0,0 +1,180 @@
...
+Right now the kpt pipeline is used to execute the conditional dance
+
+## see also
+
+- [golang](https://go.dev)
+- [configuration as data](https://github.com/GoogleContainerTools/kpt/blob/main/docs/design-docs/06-config-as-data.md)
+- [cad concepts](https://kpt.dev/book/02-concepts/)
+- [kpt](https://kpt.dev/book/02-concepts/03-functions)
+- [porch](https://kpt.dev/guides/porch-user-guide)
+- [kpt fn sdk](https://github.com/GoogleContainerTools/kpt-functions-sdk)
```
gvbalaji, line 180: Thanks for the write up. I think it will be useful to explain this with an example ( not code but in text).  Just one or two paragraphs. Let's say
for interface resource,
1. For 
2. Own
3. Watch
resources, and 
1. how the Interface function registers with the SDK
2. What Call back functions need to be implemented and how they are called
3. How eventually the interface function generated the IP/VLAN/ reources and NAD Condition working with the SDK

gvbalaji (commented):
Reviewed the design doc . Will continue with the code.

kispaljr:
This is a small bugfix: https://github.com/nokia/nephio-nephio/pull/21

krm-functions/lib/condkptsdk/sdk_watches.go
```diff
@@ -0,0 +1,38 @@
...
+// so they dont have to parse the complete resourcelist
+// Also it provide readiness feedback when an error is returned
+func (r *sdk) callGlobalWatches() {
+	for _, resCtx := range r.inv.get(watchGVKKind, []corev1.ObjectReference{}) {
+		fn.Logf("run watch: %v\n", resCtx.existingResource)
+		if resCtx.gvkKindCtx.callbackFn != nil {
+			if err := resCtx.gvkKindCtx.callbackFn(resCtx.existingResource); err != nil {
+				fn.Logf("populatechildren not ready: watch callback failed: %v\n", err.Error())
+				r.rl.Results = append(r.rl.Results, fn.ErrorConfigObjectResult(err, resCtx.existingResource))
+				r.ready = false
```
kispaljr, line 34: if the watch callback returns with an error, then this .r.ready flag is being set, but it doesn't seem to be used anywhere. 

populateCallback is being called anyway, shouldn't we prevent that?
As a user of the SDK I intuitively expected the KRM function to be stopped and return with a fail result, if I return with and error from any of the callbacks.

krm-functions/lib/condkptsdk/sdk.go
```diff
@@ -0,0 +1,113 @@
...
+
+type sdk struct {
+	cfg   *Config
+	inv   Inventory
+	rl    *fn.ResourceList
+	kptf  kptfilelibv1.KptFile
+	ready bool
+}
+
+func (r *sdk) Run() {
```
kispaljr, line 83: sdk.Run() should return with a (bool, error) so that it can be passed as a return value to fn.Run().

Note: sdk.Run() is called from fn.Run(). Since sdk.Run() never returns anything, fn.Run() has no other choice, but always returning with a success, even if an error happened in sdk.Run()

krm-functions/lib/condkptsdk/design.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Owns KRM resource
+
+The `Owns` resource filter identifies which KRM resources are children of the `for` resource instance. You could also say these are created or lifecycled as a result of the parent resource (the `for` resource in this case)
+The sdk defines different types of own resources:
+- childRemote: 
+    - the current/parent fn/controller defines the spec attributes of the child resource, but another child function/controller takes care of the actuation that are related to this KRM resource. Like updating the status and are deriving other child resources acting as a parent.
+    - A remote function will act upon this KRM through a `for` filter and will update the status. 
+    - The deletion is take care of by the remote resource by acting on the delete annotiation set by the sdk
+    - An example use case is e.g. the interface-fn that needs an IP. The interface-fn is the parent that creates an IPAllocation on which a downstream function/controller acts and fills out the
```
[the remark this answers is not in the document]
henderiw: fixed

krm-functions/lib/condkptsdk/inventory.go
```diff
@@ -0,0 +1,136 @@
...
+const (
+	actionCreate action = "create"
+	actionDelete action = "delete"
+	actionUpdate action = "update"
+	actionGet    action = "get"
+)
+
+// initializeGVKInventory initializes the GVK with the generic GVK
+// resources as specified in the SDKConfig
+// used to provide faster loopup if the GVK is relevant for the fn/controller
```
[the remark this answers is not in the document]
henderiw: fixed

krm-functions/lib/condkptsdk/sdk_updates.go
```diff
@@ -0,0 +1,170 @@
...
+func (r *sdk) deleteObjFromResourceList(obj *fn.KubeObject) {
+	for idx, o := range r.rl.Items {
+		if isGVKNNEqual(o, obj) {
+			r.rl.Items = append(r.rl.Items[:idx], r.rl.Items[idx+1:]...)
+		}
+	}
+}
+
+// isGVKNEqual validates if the APIVersion, Kind, Name and Namespace of both fn.KubeObject are equal
+func isGVKNNEqual(curobj, newobj *fn.KubeObject) bool {
```
gvbalaji, line 162: Probably a good candidate for a base library

gvbalaji (commented):
Trying to digest the end to end flow.  Will continue.

krm-functions/lib/condkptsdk/inventory.go
```diff
@@ -0,0 +1,136 @@
...
+package condkptsdk
+
+import (
+	"fmt"
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type Inventory interface {
```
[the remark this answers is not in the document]
henderiw: made internal

krm-functions/lib/condkptsdk/sdk_updates.go
henderiw: I added in utils for now

krm-functions/lib/condkptsdk/sdk.go
henderiw: I changed the implementation to bring the errors outside.

krm-functions/lib/condkptsdk/sdk_watches.go
henderiw: fixed

krm-functions/lib/condkptsdk/design.md
henderiw: added examples

krm-functions/lib/condkptsdk/sdk_populate_inventory.go
```diff
@@ -0,0 +1,146 @@
...
+	// we assume the kpt file is always resource idx 0 in the resourcelist, the object is used
+	// as a reference to errors when we encounter issues with the condition processing
+	// since conditions are stored in the kptFile
+	//o := r.rl.Items.GetRootKptfile()
+
+	// We first run through the conditions to check if an ownRef is associated
+	// to the for resource objects. We call this the forOwnerRef
+	// When a forOwnerRef exists it is used to associate a watch resource to the
+	// inventory specific to the for resource or globally.
+	for _, c := range r.kptf.GetConditions() {
```
aakashchan, line 45: This is my understanding, please correct me if I'm wrong.
Populate inventory method loops over all the condition and resources. It knows "which is the ownerRef" based on if conditions are present. 
For example in NFDeployment scenario, will the condition be present in the Kptfile? If not, then that resource relationship will not be tracked in the inventory to prepare ready context to call GenerateResourceFn. 

If the condition will be present? Which upstream function creates it?

krm-functions/lib/condkptsdk/sdk.go
```diff
@@ -0,0 +1,131 @@
...
+
+import (
+	"fmt"
+
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	kptfilelibv1 "github.com/nephio-project/nephio/krm-functions/lib/kptfile/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const FnRuntimeOwner = "fnruntime.nephio.org/owner"
```
kispaljr, line 27: I propose to remove the "fnruntime" string from teh  "fnruntime.nephio.org/owner" annotation name, since the ownership relationship between kpt package resources is independent of what SDK we use for implementing the KRM functions. 
I.e. this annotation will be used also in Specializers and 3rd party KRM functions that might not use our SDK (might not even be implemented in Go).

krm-functions/lib/condkptsdk/sdk.go
```diff
@@ -0,0 +1,131 @@
...
+const FnRuntimeDelete = "fnruntime.nephio.org/delete"
```
kispaljr, line 28: I propose to remove the "fnruntime" string from the  "fnruntime.nephio.org/delete" annotation name, since the deletion flag mechanism should work independently of what SDK we use for implementing the KRM functions. 
I.e. this annotation will be used also in Specializers and 3rd party KRM functions that might not use our SDK (might not even be implemented in Go).

kispaljr (changes requested):

krm-functions/lib/condkptsdk/inventory_diff_test.go
```diff
@@ -0,0 +1,543 @@
...
+		if len(diff.deleteObjs) != 2 {
+			t.Errorf("TestDiffWithSpecToDelete: -want 2, +got:\n%v", len(diff.deleteObjs))
+		}
+		if len(diff.updateObjs) != 0 {
+			t.Errorf("TestDiffWithSpecToUpdate: -want 0, +got:\n%v", len(diff.updateObjs))
+		}
+	}
+}
+
+/*
```
johnbelamaric, line 380: Why are these commented out? Are they no longer relevant or just the functionality is not yet implemented? Either remove them or provide a comment/TODO on how they will be resolved.

krm-functions/lib/condkptsdk/sdk.go
henderiw: or specializer.nephio.org ?

henderiw: changed to specializer.nephio.org for now

krm-functions/lib/condkptsdk/sdk.go
henderiw: see above

krm-functions/lib/condkptsdk/design.md
```diff
@@ -0,0 +1,403 @@
...
+type GenerateResourceFn func(*fn.KubeObject, fn.KubeObjects) (*fn.KubeObject, error)
+```
+
+Any fn/controller MUST implement the `GenerateResourceFn`.
+
+### sdk phases
+
+The SDK operates in phases when being executed within a fn/controller
+- firstly the sdk builds up an inventory based on the filters
+- Secondly the global watch callbacks are called. The fn/controller implenmting these callback use the data for 2 things:
```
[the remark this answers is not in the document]
henderiw: fixed

krm-functions/lib/condkptsdk/sdk.go
```diff
@@ -0,0 +1,131 @@
...
+const (
+	// ChildRemoteCondition defines a GVK resource for which only conditions need to be created
+	ChildRemoteCondition ResourceKind = "remoteCondition"
+	// ChildRemote defines a GVK resource for which conditions and resources need to be created
+	ChildRemote ResourceKind = "remote"
+	// ChildLocal defines a GVK resource for which no conditions need to be created
+	ChildLocal ResourceKind = "local"
+)
+
+type Config struct {
```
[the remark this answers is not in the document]
henderiw: I have tried to do this before but since you execute in the pkgname of the sdk I feel that it is ok to use Config since you configure the SDK. Thoughts?

krm-functions/lib/condkptsdk/sdk.go
```diff
@@ -0,0 +1,131 @@
...
+	cfg   *Config
+	inv   inventory
+	rl    *fn.ResourceList
+	kptf  kptfilelibv1.KptFile
+	ready bool
+}
+
+func (r *sdk) Run() (bool, error) {
+	if r.rl.Items.Len() == 0 {
+		r.rl.Results = append(r.rl.Results, fn.ErrorResult(fmt.Errorf("no resources present in the resourcelist")))
```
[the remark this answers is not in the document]
henderiw: If I look at the sdk it should never get to us. Will change to no-op.

krm-functions/lib/condkptsdk/sdk.go
```diff
@@ -0,0 +1,131 @@
...
+		r.rl.Results = append(r.rl.Results, fn.ErrorResult(err))
+		return false, err
+	}
+
+	// initialize inventory
+	if err := r.populateInventory(); err != nil {
+		return false, err
+	}
+	// list the result of inventory -> used for debug only
+	r.listInventory()
```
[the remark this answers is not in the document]
henderiw: great idea. Added it. One thing is that there could be multiple forInstances. Right now if it is set to 1 for instance the debug is on for all.

krm-functions/lib/condkptsdk/inventory_diff_test.go
henderiw: this was left over as we had to refactor them. resolved now

krm-functions/lib/condkptsdk/sdk_populate_inventory.go
```diff
@@ -0,0 +1,146 @@
...
+		fn.Logf("set existing object in inventory, kind %s, ref: %v ownerRef: %v\n", gvkKindCtx.gvkKind, ref, nil)
+		if err := r.inv.set(gvkKindCtx, []corev1.ObjectReference{*ref}, x, false); err != nil {
+			fn.Logf("error setting exisiting object in the inventory: %v\n", err.Error())
+			r.rl.Results = append(r.rl.Results, fn.ErrorConfigObjectResult(err, o))
+			return err
+		}
+	case ownGVKKind:
+		ownerKindCtx, ok := r.inv.isGVKMatch(ownerRef)
+		if !ok || ownerKindCtx.gvkKind != forGVKKind {
+			// this means the resource was added from a different kind
```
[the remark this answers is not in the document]
henderiw: indeed. We have this case e.g. DNN generates an IP allocation (child) for a pool and Interface generates IP allocation children for itself. So this case deals with this.

krm-functions/lib/condkptsdk/sdk.go
```diff
@@ -0,0 +1,131 @@
...
+	ChildRemoteCondition ResourceKind = "remoteCondition"
+	// ChildRemote defines a GVK resource for which conditions and resources need to be created
+	ChildRemote ResourceKind = "remote"
+	// ChildLocal defines a GVK resource for which no conditions need to be created
+	ChildLocal ResourceKind = "local"
+)
+
+type Config struct {
+	For                    corev1.ObjectReference
+	Owns                   map[corev1.ObjectReference]ResourceKind    // ResourceKind distinguishes ResourceKindNone and ResourceKindFull
```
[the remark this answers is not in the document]
henderiw: yes changed

krm-functions/lib/condkptsdk/sdk_populate_inventory.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (r *sdk) populate(forOwnerRef, ref, ownerRef *corev1.ObjectReference, x any, o *fn.KubeObject) error {
+	// we lookup in the GVK context we initialized in the beginning to validate
+	// if the gvk is relevant for this fn/controller
+	// what the gvk Kind is about through the kindContext
+	gvkKindCtx, ok := r.inv.isGVKMatch(getGVKRefFromGVKNref(ref))
+	if !ok {
+		// it can be that a resource in the kpt package is not relevant for this fn/controller
+		// As such we return
+		fn.Logf("populate no match, ref: %v \n", ref)
```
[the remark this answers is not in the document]
henderiw: I put it behind the debug flag

krm-functions/lib/condkptsdk/sdk_populate_inventory.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+	}
+	// Now we have the forOwnerRef we run through the condition again to populate the remaining
+	// resources in the inventory
+	for _, c := range r.kptf.GetConditions() {
+		ref := kptfilelibv1.GetGVKNFromConditionType(c.Type)
+		ownerRef := kptfilelibv1.GetGVKNFromConditionType(c.Reason)
+		x := c
+		if err := r.populate(forOwnerRef, ref, ownerRef, &x, r.rl.Items.GetRootKptfile()); err != nil {
```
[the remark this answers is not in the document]
henderiw: yes updated to use the kptfile in the sdk

krm-functions/lib/condkptsdk/sdk.go
johnbelamaric: sgtm

krm-functions/lib/condkptsdk/sdk.go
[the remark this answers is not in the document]
johnbelamaric: sure, that's fine

PR #128, opened 2023-04-21
Avoid YAML marshaling round-trips and use existing upstream SDK functions

kispaljr:
SDK functions will utilize the already parsed YAML nodes inside the KubeObject, and thus they are more efficient in case of GetGoStruct().
Otherwise I propose to reuse upstream SDK functions, instead of reimplementing them.


gvbalaji (approved):
/lgtm

PR #129, opened 2023-04-21
added getters into kubeobject

henderiw:
added getters to kubeobject

### DIFF
--- a/krm-functions/lib/kubeobject/kubeobject.go
+++ b/krm-functions/lib/kubeobject/kubeobject.go
@@ -32,6 +32,50 @@ func (r *KubeObjectExt[T1]) GetGoStruct() (T1, error) {
 	return x, err
 }
 
+func (r *KubeObjectExt[T1]) GetNestedString(fields ...string) string {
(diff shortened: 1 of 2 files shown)


johnbelamaric (commented):
Do we think it's safe to ignore these errors? What could be the cause of these errors? Malformed YAML?

gvbalaji (commented):
Thanks @henderiw .  Now that we refactored and just extending the base kubeobject, not sure we need these indirection methods as we can not ignore errors. Probably we can just invoke the base methods from the methods in the individual libraries ( interface, nad etc) ?

henderiw:
we can expose the errors and present it to the user if needed.

johnbelamaric:
But wouldn't that be the same as calling the original methods?

johnbelamaric:
Oh, I see it would go from 3 to 2 return values?

henderiw:
Right now this is an intermediate step  until we go to full go struct approach. We need to validate if the gostruct approach works in all circumstances.

PR #134, opened 2023-04-26
Add helper functions for testing KRM functions

kispaljr:
Moving the helper functions that proved to be useful while testing dnn-fn to a common test lib


gvbalaji (approved):
Thanks @kispaljr . Looks good.

PR #135, opened 2023-04-27
sdk nil pointer. check missing

henderiw:
fixed a missing nil pointer ref check


gvbalaji (commented):
Thanks @henderiw . Looks like this includes some other commits as well (getters for Kubeobject). Can you please check?

henderiw:
@gvbalaji strange it came in as I was changing this from base, but reverted. Thx

ganchandrasekaran (approved):

PR #140, opened 2023-04-28
VLAN alloc with new KO

henderiw:
this PR implements the VLAN Allocation library with the new KubeObject aligned with SetSpec/SetStatus


krm-functions/lib/vlanalloc/v1alpha1/vlanalloc.go
```diff
@@ -0,0 +1,57 @@
...
+
+package v1alpha1
+
+import (
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	"github.com/nephio-project/nephio/krm-functions/lib/kubeobject"
+	vlanv1alpha1 "github.com/nokia/k8s-ipam/apis/alloc/vlan/v1alpha1"
+)
+
+type VLANAllocation struct {
```
gvbalaji, line 25: Is there a need to create this structure? Can the functions developers  just directly create the base KubeObjectExt structure?  All the methods are used from there anyway.

krm-functions/lib/vlanalloc/v1alpha1/vlanalloc.go
```diff
@@ -0,0 +1,57 @@
...
+	kubeobject.KubeObjectExt[*vlanv1alpha1.VLANAllocation]
+}
+
+// NewFromKubeObject creates a new KubeObjectExt
+// It expects a *fn.KubeObject as input representing the serialized yaml file
+func NewFromKubeObject(o *fn.KubeObject) (*VLANAllocation, error) {
```
gvbalaji, line 31: In the same spirit as above comment, show we just return the base KubeObjectExt from here? 
In fact I am also thinking can the function devs directly invoke kubeobject.NewFromKubeObject[*vlanv1alpha1.VLANAllocation](fn.KubeObject)?

That could mean do we really need this library for indirection?

gvbalaji (commented):
Thanks @henderiw . I am on the fence if we realy need this with all the changes we made to the base KubeObjectExt. Added a couple of comments.

krm-functions/lib/vlanalloc/v1alpha1/vlanalloc.go
henderiw: The value is you get explicit type checking for the object and we need this in several libraries. This is the reason to do this. otherwise people need to implement this in each function independently again and again.

krm-functions/lib/vlanalloc/v1alpha1/vlanalloc.go
henderiw: The value is you get explicit type checking for the object and we need this in several libraries. This is the reason to do this. otherwise people need to implement this in each function independently again and again.

gvbalaji (approved):
/approve

krm-functions/lib/vlanalloc/v1alpha1/vlanalloc.go
gvbalaji: Thanks @henderiw . Approved.

krm-functions/lib/vlanalloc/v1alpha1/vlanalloc.go
```diff
@@ -0,0 +1,57 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	"github.com/nephio-project/nephio/krm-functions/lib/kubeobject"
+	vlanv1alpha1 "github.com/nokia/k8s-ipam/apis/alloc/vlan/v1alpha1"
```
vjayaramrh, line 22: curious, will this always be in the nokia repo?

henderiw: no this is to be merged. But this is the workaround for now. So next week we start bringing this into nephio.

kispaljr:
I understand now, that the main concern is the lack of static type checking in the `interface{}` type parameter of `SetSpec` and `SetStatus`. I agree that this is a valid concern, however I suggest to use the power of Go generics to fix this, as opposed to implementing it one-by-one for each type (avoiding this is the main reason for having generics). 
Here s my proposal to add static type checking: https://github.com/nephio-project/nephio/pull/141

PR #141, opened 2023-04-28
Add static type checking for KubeObjectExt methods

kispaljr:
@henderiw was concerned with the lack of static type checking in KubeObjectExt, and I agree. This is what this PR tries to fix by adding static type checking to NewFromGoStruct, SetSpec and SetStatus.

For backward compatibility reasons, the original SetSpec and SetStatus methods were kept, and the new methods were added as SafeSetSpec and SafeSetStatus.

### DIFF
--- a/krm-functions/lib/kubeobject/kubeobject.go
+++ b/krm-functions/lib/kubeobject/kubeobject.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"reflect"
 
 	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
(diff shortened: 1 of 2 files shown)


gvbalaji (approved):
Nice. thanks @kispaljr 
/approve

gvbalaji:
Thanks @kispaljr . Can you please only have kubeobjectExt changes in  this PR and have separate ones for interface and VLAN? I have a feeling if we are happy with kubeobjectExt, we may not even need the others as kubeobjectExt will provide enough type safety.

kispaljr:
> Thanks @kispaljr . Can you please only have kubeobjectExt changes in this PR and have separate ones for interface and VLAN? I have a feeling if we are happy with kubeobjectExt, we may not even need the others as kubeobjectExt will provide enough type safety.

Well, I was able to do something similar (the commit is already in the PR): 
I reverted the changes to those libs.
I also kept the original SetSpec/SetStatus methods, and renamed the new type-checked ones to SafeSetSpec/SafeSetStatus, thus keeping the KubeObjectExt API backward compatible. This is needed so that the already merged Interface and VLAN libs will compile on this branch, otherwise most of our presubmit test would fail for this PR, preventing it to be merged.

If we decide to refactor the libs later, we can rename SafeSetSpec back to SetSpec.

kispaljr:
Proposal was rebased onto the new KubeObjectExt implementation.
In its current form this PR is backward-compatible, and only:
- adds type safety to NewFromGoStruct, and 
- adds 2 new methods: SafeSetSpec and SafeSetStatus

gvbalaji (approved):
Thanks @kispaljr .
/approve

PR #146, opened 2023-05-01
NFDeployment newKO

aakashchan:
aligned the interface with the new parser implementation
added setters using the real types as this eases validation.

Unit tested for UPFDeployment type. I will add more tests SMF and AMF type in subsequent PRs if needed.

### DIFF
--- a/krm-functions/lib/nfdeploy/v1alpha1/nfdeployment.go
+++ b/krm-functions/lib/nfdeploy/v1alpha1/nfdeployment.go
@@ -1,0 +1,67 @@
+/*
+ Copyright 2023 The Nephio Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	nephiodeployv1alpha1 "github.com/nephio-project/api/nf_deployments/v1alpha1"
+	"github.com/nephio-project/nephio/krm-functions/lib/kubeobject"
+)
+
+type NfDeploymentType interface {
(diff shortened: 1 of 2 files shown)


gvbalaji (commented):
Thanks @aakashchan . Looks good but I have a question. With the type safety provided at the base `KubeObjectExt` is this layer really needed? Can the functions directly call the base library with the required type safety?

henderiw:
The differences between the approaches are very small.
Here is my breakdown:
- KubeObjectExt: expects the developer to know generics and add the right type in the generic field of the go struct. More error prune, but checked in the library.
- Explicit typed interface. needs explicit code although could be generated. Easier for people to grasp.

So it depends on the target for these. A deep go developer prefers KubeObjectExt, the less experienced dev prefers Explicit typed interfaces. How I personally look at this is I look at the consumer of this and given the SDK targets less experienced people. As such this is why we created the explicit methods.

Personally I prefer KubeObjectExt, but the barrier for the consumers is higher.
My 2 cents.

Example Here.

Explicit
			alloc, err := vlanlibv1alpha1.NewFromKubeObject(vlanalloc)
			if err != nil {
				return nil, err
			}
KubeObjectExt
			alloc, err := ko.NewFromKubeObject[*vlanv1alpha1.VLANAllocation](vlanalloc)
			if err != nil {
				return nil, err
			}
			
@aakashchan what is your view?

aakashchan:
@henderiw @gvbalaji 

1. Currently, the nf deploy lib already expects user to know about this generic type. So we are not obliterating generics from user completely. I don't see the point of repeating the code in library without generics for all 3 three types.

```
type NfDeploymentType interface {
	*nephiodeployv1alpha1.UPFDeployment | *nephiodeployv1alpha1.SMFDeployment | *nephiodeployv1alpha1.AMFDeployment
}
```
2. Another draw back is, SetSpec() or SetStatus() expects argument of `NFDeploymentSpec` than  `UPFDeployment`. Here, library couldn't do much even with generics.

3. Bright side, It ensures tighter type safety on one hand. 

```
i, err := NewFromYAML[*nephiodeployv1alpha1.UPFDeployment]([]byte(tc.file))
g, err := i.GetGoStruct()
// error check ...
```

I also prefer using KubeObjExt directly over wrapper lib.

henderiw:
if you do this we do not expose generics to the user.
// NewFromKubeObject creates a new parser interface
// It expects a *fn.KubeObject as input representing the serialized yaml file
func NewFromKubeObject(o *fn.KubeObject) (*Interface, error) {
	r, err := kubeobject.NewFromKubeObject[*nephioreqv1alpha1.Interface](o)
	if err != nil {
		return nil, err
	}
	return &Interface{*r}, nil
}

henderiw:
Also Isti made the SetSpec and SetStatus such that they can also use the generic type. But this PR is not merged yet

aakashchan:
Closing this as we decided to use Kubeobject ext directly

PR #148, opened 2023-05-01
KubeobjectExt should keep formatting of lists

kispaljr:
Rewrite KubeObjectExt setters to keep formatting of lists, not just structs and maps


gvbalaji:
Thanks @kispaljr . I see a  few TODOs. Are they for later ?

kispaljr:
> Thanks @kispaljr . I see a few TODOs. Are they for later ?

Yes, you are right to notice this: I kept working on the PR even after submitting to review, sorry.
By now I've completed all TODOs and refactoring that I planned to do, except for 1 TODO, that we can safely leave for later. (That will most probably fixed only when I submit the whole thing into the underlying SDK).
All-in-all I try not to add any commits from now on, except if a reviewer asks.

PR #164, opened 2023-05-05
Make presubmit tests to run on all go modules in the repo

kispaljr:


radoslawc:
> But I am not familiar with prow usage patterns, so this may be a completely wrong direction
There are no good or bad approaches here I feel. As long as it's working and it's comfortable for daily use it's awesome. That being said I like that approach from practical point of view having fairly uniform .prow.yaml file between the projects and adjust things in Makefiles (that will obviously differ from project to project anyway).

run-for-all-go-modules.sh
johnbelamaric, line 1: missing newline at the end

Makefile
```diff
@@ -25,29 +25,31 @@ endif
 unit: ## Run unit tests against code.
 ifeq ($(CONTAINER_RUNNABLE), 0)
 		$(CONTAINER_RUNTIME) run -it -v ${PWD}:/go/src -w /go/src docker.io/library/golang:${GO_VERSION}-alpine3.17 \
-         /bin/sh -c "go test ./... -v -coverprofile ${TEST_COVERAGE_FILE}; \
+         sh -e ./run-for-all-go-modules.sh "go test ./... -v -coverprofile ${TEST_COVERAGE_FILE}; \
```
johnbelamaric, line 28: More typical approach I have seen is a list of subdirs in each Makefile and then looping through that and calling `make test` on each. But this is fine too. Also seems like there is probably a more "make native" way, but this works.

kispaljr: I am not a Makefile guru either, so I am pretty sure that there is a more make-native way :)

If all of my higher priority tasks are solved, I will try to put the list of subdirs into the Makefile and call make recursively, as you've suggested. I would still prefer to assemble the list dynamically (by finding go.mod files) as opposed to have a manually maintained list

PR #169, opened 2023-05-06
Specialzer

henderiw:
This is the PR for the specialisers: IPAM/VLAN

They use the condSDK and are executed in a controller as a local fn. This allows to reuse a lot of code and second to use the IPAM/VLAN fns as stubs in the test pipelines.


krm-functions/ipam-fn/fn/function.go
```diff
@@ -0,0 +1,84 @@
...
+	fn.Logf("ipalloc: %v\n", forObj)
+	allocKOE, err := kubeobject.NewFromKubeObject[ipamv1alpha1.IPAllocation](forObj)
+	if err != nil {
+		return nil, err
+	}
+	alloc, err := allocKOE.GetGoStruct()
+	if err != nil {
+		return nil, err
+	}
+	resp, err := r.ClientProxy.Allocate(context.Background(), alloc, nil)
```
kispaljr, line 69: If I understand correctly this is intended to use both with a mock backend proxy, but also with an actual backend, as well. The later case seems to contradict with the "lambda function" nature of KRM functions: it may generate different results for the same input.
I understand why this function is useful for testing purposes, but for calling out to the real backend, shouldn't we  use a Kubernetes Operator, more precisely a Specializer instead?

kispaljr: My understanding is that the Specializer should watch the kpt package and detect xxxAllocations with an empty `status` automatically, without the help of a KRM function. Isn't it the case?

henderiw: this is the call to the Grpc call to the backend. As long as the backend does not get updated the result will be the same. So I don't see the contradiction.

henderiw: I exposed the fn because we can run them independently for our porch independent test pipeline. But in a real environment this will run as a specialised controller and not through the kpt pipeline.

kispaljr: yes, that was my question, thanks for clarifying.

PR #171, opened 2023-05-08
ipam-fn

henderiw:
splitter the specialised in functional blocks to be able to merge and run positive tests


kispaljr:
It's worth mentioning that this KRM function is not meant to be part of a kpt package's pipeline, but will be part of a Specializer's implementation (and also used for testing)

Issue #165, opened 2023-05-05
Package specialization - Integration test framework


gvbalaji:
Develop the integration test framework for package specialization which can help test the wntite kpt pipeline to transform input KRM resources to the actual output  KRM resources that be applied to workload clusters specifically UPFDeployment, AMFDeployment AND SMFDeployment.

henderiw:
work in progress, almost ready but dependent on the fn PR(s).

kispaljr:
Progress is tracked here: https://github.com/nephio-project/nephio/pull/178

kispaljr:
@gvbalaji this was closed by accident. 
There are two PRs related to this, and only one of them was merged. Apparently this was enough for PROW to close this. I am reopening this, until the other PR is also merged.

PR #178, opened 2023-05-09
Pkg Specialization Pipeline tests

kispaljr:
Add a "middle-sized test" that can test a chain/pipeline of multiple KRM functions end-to-end, but doesn't need a Kubernetes cluster, because it is built on the built-in Go testing framework.
This is less than a real integration or end-to-end test, but more than a unit test, even more than the component tests of single KRM function.

Fixes https://github.com/nephio-project/nephio/issues/165


gvbalaji (commented):
Thanks @kispaljr . Should dnn function be part of this PR?  I was thinking it may be in a different PR already.

aakashchan (approved):
lgtm

gvbalaji:
Thanks @kispaljr . Looks like tests are failing for this PR. Can you PTAL?

PR #179, opened 2023-05-09
reworked repo to allow for mix of specializer controller and regular …

henderiw:
I reworked the repo a bit to make it easier to contribute controllers and specializers


controllers/pkg/specializerreconciler/condition.go
```diff
@@ -14,7 +14,7 @@
  limitations under the License.
 */
 
-package reconciler
+package specializerreconciler
```
gvbalaji, line 17: NIt. should we rename it to specializer-reconciler?

gvbalaji (approved):
Added a comment on renaming a folder ( minor). Else looks good. 
/approve

controllers/pkg/specializerreconciler/condition.go
henderiw: done

vjayaramrh (approved):

PR #181, opened 2023-05-10
controller helpers

henderiw:
this is a set of controller helpers to make it easier to build controllers

### DIFF
--- a/controllers/pkg/resource/api.go
+++ b/controllers/pkg/resource/api.go
@@ -1,0 +1,190 @@
+/*
+Copyright 2023 The Nephio Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resource
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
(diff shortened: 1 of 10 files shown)


controllers/pkg/resource/api.go
```diff
@@ -0,0 +1,183 @@
...
+// NewAPIPatchingApplicator returns an Applicator that applies changes to an
+// object by either creating or patching it in a Kubernetes API server.
+func NewAPIPatchingApplicator(c client.Client) APIPatchingApplicator {
+	return APIPatchingApplicator{c}
+}
+
+// Apply changes to the supplied object. The object will be created if it does
+// not exist, or patched if it does. If the object does exist, it will only be
+// patched if the passed object has the same or an empty resource version.
+func (a *APIPatchingApplicator) Apply(ctx context.Context, o client.Object, ao ...ApplyOption) error {
```
aakashchan, line 46: `sigs.k8s.io/cluster-api/util/patch` package have PatchHelper struct. Is there a reason why we don't want to use it?

aakashchan:
Do we need a helper on the controller-runtime utils for Finalizer and Patching?

controllers/pkg/resource/api.go
henderiw: The whole idea is to present the client the same semantics as the regular client. E.g. Apply and Create  for the developer have the exact same semantics. So under the hood patch is used.

if err := r.Apply(ctx, secret); err != nil {
			cr.SetConditions(infrav1alpha1.Failed(err.Error()))
			r.l.Error(err, "cannot create secret")
			return err
		}

henderiw:
eventually I would like to appear in controller runtime, but I did not find it anywhere so far in a way that it extends the client gracefully.

gvbalaji:
This may have been covered already by the bootstrap controller PR? @henderiw

henderiw:
this is covered by the bootstrap controller now

PR #194, opened 2023-05-15
Return err on `condsdk.New` returns error

subh007:
Return err on `condsdk.New` returns error on setting up sdk. Also avoid further execution on error.

### DIFF
--- a/krm-functions/dnn-fn/fn/function.go
+++ b/krm-functions/dnn-fn/fn/function.go
@@ -74,7 +74,7 @@ func Run(rl *fn.ResourceList) (bool, error) {
 	)
 	if err != nil {
 		rl.Results.ErrorE(err)
-		return false, nil
+		return false, err
 	}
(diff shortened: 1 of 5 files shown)


gvbalaji (approved):
/approve

kispaljr:
Just to clarify for others: 
- according to the current code in case of an SDK initialization error in an KRM function we are not stopping the whole pipeline, just make a note of the error in the output of the function (and thus the whole pipeline) and let the rest of the pipeline to run.
- according to this proposal we should stop executing the whole pipeline if SDK initialization has failed in any KRM function

IMHO: SDK initialization errors are mostly due to logical errors in the implementation (a mistake made by the developer)  and not input validation errors (a mistake made by the kpt package creator), so it is justified to halt the execution of the pipeline in these cases.


/lgtm

gvbalaji:
@johnbelamaric @henderiw PTAL and approve to merge this.

PR #198, opened 2023-05-17
repo-token controllers

henderiw:
PR that covers the implementation of the repo and token reconciler.


controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"code.gitea.io/sdk/gitea"
+	"github.com/go-logr/logr"
+	ctrlconfig "github.com/henderiw-nephio/nephio-controllers/controllers/config"
+	"github.com/henderiw-nephio/nephio-controllers/pkg/applicator"
```
johnbelamaric, line 27: Should these henderiw-nephio packages move to nephio-project?

controllers/pkg/reconcilers/repository/README.md
```diff
@@ -0,0 +1,39 @@
+# repository controller
+
+The repo controller is a k8s controller acting on repository.infra.nephio.org and handles the lifecycle of the repository in gitea.
```
johnbelamaric, line 3: Post R1, we should consider separating consumption/requirements/provisioning resources.

controllers/pkg/reconcilers/repository/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+For each repo CR the repo-controller handles the lifecycle of the repository in gitea. Updates are limited based on gitea's capabilities, so it is better to delete and recreate the repo or handle updates directly in gitea.
+
+## implementation
+
+The implementation assumes the repo-controller runs in the same cluster as the gitea server. Based on the environment variables we help the controller to connect to the gitea server.
```
johnbelamaric, line 9: Post R1, we should consider Gitea running anywhere. Overall we want the model to work with Gitea, GitHub, GitLab, and other Git providers.

controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+	infrav1alpha1 "github.com/nephio-project/api/infra/v1alpha1"
+	"github.com/nephio-project/nephio/controllers/pkg/giteaclient"
+	"github.com/nokia/k8s-ipam/pkg/meta"
+	"github.com/nokia/k8s-ipam/pkg/resource"
```
johnbelamaric, line 31: Are these general purpose packages? Should they move to nephio-project?

controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.l = log.FromContext(ctx)
+	r.l.Info("reconcile", "req", req)
+
+	cr := &infrav1alpha1.Repository{}
+	if err := r.Get(ctx, req.NamespacedName, cr); err != nil {
+		// if the resource no longer exists the reconcile loop is done
+		if resource.IgnoreNotFound(err) != nil {
+			r.l.Error(err, "cannot get resource")
+			return ctrl.Result{}, errors.Wrap(resource.IgnoreNotFound(err), "cannot get resource")
```
johnbelamaric, line 88: can we just use err here instead of resource.IgnoreNotFound(err)?

controllers/pkg/reconcilers/token/README.md
```diff
@@ -0,0 +1,49 @@
+# token controller
+
+The token controller is a k8s controller acting on token.infra.nephio.org and handles the lifecycle of the token in gitea. It also adds a corresponding secret in k8s within the namespace the token was applied.
+
+The token is immutable, so if you want to change the token it has to be deleted/created
+
+## implementation
+
+The implementation assumes the repo-controller runs in the same cluster as the gitea server. Based on the environment variables we help the controller to connect to the gitea server.
```
liamfallon, line 9: The implementation assumes the **token-controller**

controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	// check if client exists otherwise retry
+	giteaClient := r.giteaClient.Get()
+	if giteaClient == nil {
+		err := fmt.Errorf("gitea server unreachable")
+		r.l.Error(err, "cannot connect to git server")
+		cr.SetConditions(infrav1alpha1.Failed(err.Error()))
+		return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, errors.Wrap(r.Status().Update(ctx, cr), errUpdateStatus)
```
johnbelamaric, line 99: Seems pretty aggressive to retry every 5 seconds forever. I thought controller runtime automatically implements exponential backoff (this is hearsay...)?

johnbelamaric (commented):
Some comments and questions but I think they are mostly post-R1 and I would consider them non-blocking.

controllers/pkg/reconcilers/token/README.md
henderiw: yes fixed

controllers/pkg/reconcilers/repository/README.md
henderiw: agreed. there is 2 things needed for this in my view:
1. a dns setup that allows to get to a URL from anywhere. So with this we can reach the server uniformly.
2. we need to come up with a means to represent the gift provider. There is multiple options here.
- introduce a type in the CR ( I have the feel the git semantics are uniform)
- introduce a provider with a dedicated implementation
- other?

controllers/pkg/reconcilers/repository/reconciler.go
henderiw: I believe there is a different behaviour when the resource is not there with not found versus resource cannot be retrieved. In the ignore not found case it stops. in the other case it triggers a new reconcile.

controllers/pkg/reconcilers/repository/reconciler.go
henderiw: yes I can just requeue.

henderiw: changed

controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+			cr.SetConditions(infrav1alpha1.Failed(err.Error()))
+			return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, errors.Wrap(r.Status().Update(ctx, cr), errUpdateStatus)
+		}
+
+		r.l.Info("Successfully deleted resource")
+		return ctrl.Result{Requeue: false}, nil
+	}
+
+	// add finalizer to avoid deleting the repo w/o it being deleted from the git server
+	if err := r.finalizer.AddFinalizer(ctx, cr); err != nil {
```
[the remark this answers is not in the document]
henderiw: agreed, I took a shortcut for now.

controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+		r.l.Error(err, "cannot add finalizer")
+		cr.SetConditions(infrav1alpha1.Failed(err.Error()))
+		return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, errors.Wrap(r.Status().Update(ctx, cr), errUpdateStatus)
+	}
+
+	// create repo in git server
+	if err := r.createRepo(ctx, giteaClient, cr); err != nil {
```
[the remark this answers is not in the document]
henderiw: ok I added an update handling for the parameters it supports. if the update fails we will go to non ready state until we resolve it.

before the update was not done on the repo. Now we update the parameters it allows to update.

controllers/pkg/reconcilers/token/README.md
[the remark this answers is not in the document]
henderiw: in the manifest packages there is a separate yaml for this. I split token and repo since token is independent on repo and we need to deal with different way's configsync/porch handles them, so this Is why I split and in the repo package we also have tokens we will apply to the cluster.

controllers/pkg/reconcilers/repository/README.md
henderiw: agreed we should reflect on the purpose of these and define the right strategy behind each one.

controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -0,0 +1,229 @@
...
+		return ctrl.Result{Requeue: true}, errors.Wrap(r.Status().Update(ctx, cr), errUpdateStatus)
+	}
+
+	if meta.WasDeleted(cr) {
+		// repo being deleted
+		// Delete the repo from the git server
+		// when successfull remove the finalizer
+		if err := r.deleteRepo(ctx, giteaClient, cr); err != nil {
+			r.l.Error(err, "cannot delete repo in git server")
+			return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, errors.Wrap(r.Status().Update(ctx, cr), errUpdateStatus)
```
johnbelamaric, line 109: same every-five-seconds comment here and at least one other place below

controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -0,0 +1,229 @@
...
+		if err != nil {
+			r.l.Error(err, "cannot create repo")
+			// Here we dont provide the full error sicne the message change every time and this will retrigger
+			// a new reconcile loop
+			cr.SetConditions(infrav1alpha1.Failed("cannot create repo"))
+			return err
+		}
+		r.l.Info("repo created", "name", cr.GetName())
+		cr.Status.URL = &repo.CloneURL
+	} else {
```
johnbelamaric, line 186: if we `return nil` just before this else, we can get rid of the else and the indentation associated with it for the rest of the method - there is nothing else left in the method except this else clause.

johnbelamaric (commented):
Ok, great. Couple more comments, but let's hold off on merge until bootstrap merges and we can switch to the nephio imports.

controllers/pkg/reconcilers/repository/reconciler.go
henderiw: changed

PR #199, opened 2023-05-17
bootstrap controller

henderiw:
This is the bootstrap controller
- installs secrets
- installs the staging packages to the workload


controllers/pkg/meta/unstructured.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func GetUnstructuredFromGVK(gvk *schema.GroupVersionKind) *unstructured.Unstructured {
+	var u unstructured.Unstructured
+	u.SetAPIVersion(gvk.GroupVersion().String())
+	u.SetKind(gvk.Kind)
+	uCopy := u.DeepCopy()
```
johnbelamaric, line 28: nit: Why not just return &u? Or at least u.DeepCopy()?

controllers/pkg/reconcilers/bootstrap/README.md
```diff
@@ -0,0 +1,15 @@
...
+The bootstrap controller is a k8s controller which goal is to bootstrap packages on a newly installed cluster. This ensure e.g. that a gitops tool like `config-sync` is installed on a cluster and subsequent configurations tasks get handled through the gitops toolchain.
+
+## implementation
+
+The controller acts on a secret. This ensures that the bootstrap controller is agnostic to the cluster implementation. It figures out the type of cluster based on the signatures of the secret. (right now only cluster api is implemented, but the code is able to handle other implementations)
+
+The are 2 scenario's wrt secret:
+
+- A secret representing the kubeconfig of a cluster. This secret (representing the kubeconfig of the remote cluster) is used to access the cluster. The bootstrap controller install the packages in the staging repo belonging to the cluster when the cluster is ready.
+    - clustername is derive from the annotation: `nephio.org/site`
```
johnbelamaric, line 12: We haven't defined the label schema explicitly yet. In https://github.com/nephio-project/nephio-example-packages/blob/main/5g-core-topology/packagevariantset-amf-regional.yaml we picked nephio.org/site-type as being potentially "regional", "edge", "combined". Here we are defining "nephio.org/site". I think we need to standardize on this soon, if not now. Of course, additional labels and implementations may vary, but we should have some sort of standard if we are going to rely upon it to this extent.

I tend to think of site types as things like "global", "regional", "zonal" (maybe), "aggregate", "edge", "far-edge", "near-edge", and that there may be multiple clusters at any given site. So, I think something like:

* nephio.org/site-type - as above
* nephio.org/site - the name or code of a specific site (or should this be nephio.org/site-code?)
* nephio.org/region - calling out one specific site-type code as a separate thing, assuming a site hierarchy
* nephio.org/cluster-type - "workload" or "management"
* nephio.org/cluster-name - for clusters themselves, this is not needed as there is metadata.name, but we could use it for ancillary resources like these secrets

It's not perfect, but should get us pretty far, I think.

controllers/pkg/reconcilers/bootstrap/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// if the secret is being deleted dont do anything for now
+	if cr.DeletionTimestamp != nil {
+		return reconcile.Result{}, nil
+	}
+
+	// this branch handles installing the secrets to the remote cluster
+	if cr.GetNamespace() == "config-management-system" {
```
johnbelamaric, line 99: This controller is running on the management cluster, isn't it?

So, the Secret is created in the "config-management-system" namespace of the management cluster (the local cluster)? Shouldn't it be in the same namespace as the Cluster resource? I thought the CS secret just had to live in the config-management-system namespace on the remote cluster.

controllers/pkg/reconcilers/bootstrap/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+							return ctrl.Result{RequeueAfter: 10 * time.Second}, errors.Wrap(err, msg)
+						}
+					}
+				}
+			}
+			if !found {
+				return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+		}
+	} else {
```
johnbelamaric, line 152: Can we just return from the previous branch and therefore not need an "else"?

controllers/pkg/reconcilers/bootstrap/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+func (r *Reconciler) getResources(ctx context.Context, clusterName string) ([]unstructured.Unstructured, error) {
+	repos := &porchconfigv1alpha1.RepositoryList{}
+	if err := r.porchClient.List(ctx, repos); err != nil {
+		return nil, err
+	}
+
+	stagingRepoName := ""
+	for _, repo := range repos.Items {
+		if _, ok := repo.Annotations["nephio.org/staging"]; ok {
+			stagingRepoName = repo.GetName()
```
johnbelamaric, line 196: We could end up with multiple, and then with non-deterministic behavior, depending which one gets picked (is last in the list). We should either accumulate a list here, and apply everything from all, or we should use a label on the Secret to point to a specific repo, or we should just pass a specific repo name into the controller at initialization time.

controllers/pkg/reconcilers/bootstrap/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+			r.l.Error(err, "cannot get resources", "key", prKey)
+			return nil, err
+		}
+		resources = append(resources, res...)
+	}
+
+	return resources, nil
+}
+
+func includeFile(path string, match []string) bool {
```
johnbelamaric, line 230: Soon we need to build some util libs for controllers that do this, I do the same in PV, and I know we did it in other places too. Not this PR, not yet...

controllers/pkg/resource/api.go
```diff
@@ -0,0 +1,183 @@
...
+limitations under the License.
+*/
+
+package resource
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/nokia/k8s-ipam/pkg/meta"
```
johnbelamaric, line 23: Should this migrate to nephio-proejct?

controllers/pkg/resource/api.go
johnbelamaric, line 1: I am curious to learn more about all the files in this dir...

I am out of time to review them in detail right now, but didn't want to hold up my other comments, so submitting the review now.

controllers/pkg/meta/unstructured.go
henderiw: fixed

controllers/pkg/reconcilers/bootstrap/README.md
henderiw: agreed. I will use nephio.org/cluster-name for this one.

controllers/pkg/reconcilers/bootstrap/README.md
```diff
@@ -0,0 +1,15 @@
+# bootstrap controller
+
+The bootstrap controller is a k8s controller which goal is to bootstrap packages on a newly installed cluster. This ensure e.g. that a gitops tool like `config-sync` is installed on a cluster and subsequent configurations tasks get handled through the gitops toolchain.
+
+## implementation
+
+The controller acts on a secret. This ensures that the bootstrap controller is agnostic to the cluster implementation. It figures out the type of cluster based on the signatures of the secret. (right now only cluster api is implemented, but the code is able to handle other implementations)
```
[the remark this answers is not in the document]
henderiw: agreed, we need to see how other implementation deal with this going fwd. But also the secret management is a topic this is dependent upon

controllers/pkg/reconcilers/bootstrap/reconciler.go
henderiw: let me clarify:
- there is a token in the repo manifest that ask for a token for config-sync in ns config-system-management. This is applied to the management cluster but it is not used here. It will be picked up by the bootstrap controller to apply to the remote cluster.

The alternative would have been that I had to add the cluster specific handling in the token controller and I decided not to do this for not. So the code to deal with connecting to remote clusters is only in the bootstrap controller.

controllers/pkg/reconcilers/bootstrap/reconciler.go
henderiw: yes this is reused all over the place. Even in porch. So we should build some libs to interact with porch.

controllers/pkg/resource/api.go
henderiw: yes will do.

controllers/pkg/resource/api.go
henderiw: I will upload them and resolve this.

henderiw: done

controllers/pkg/resource/api.go
henderiw: done

controllers/pkg/reconcilers/bootstrap/reconciler.go
johnbelamaric: ok. I think we can go for this with R1 but I foresee issues - for example, we should support multiple namespaces which could have the same names for repos, and we could get collisions here.

controllers/pkg/reconcilers/bootstrap/reconciler.go
johnbelamaric: I think it's OK, yeah.

controllers/pkg/reconcilers/bootstrap/reconciler.go
johnbelamaric: exactly, I want to do that in porch/controllers/pkg I think.

controllers/pkg/resource/annotations.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright 2022 Nokia.
```
johnbelamaric, line 2: Nephio please

controllers/pkg/resource/annotations_test.go
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2022 Nokia.
```
johnbelamaric, line 2: Nephio please

controllers/pkg/resource/finalizer.go
```diff
@@ -0,0 +1,93 @@
+/*
+Copyright 2022 Nokia.
```
johnbelamaric, line 2: ditto

controllers/pkg/resource/finalizer_test.go
```diff
@@ -0,0 +1,190 @@
+/*
+Copyright 2022 Nokia.
```
johnbelamaric, line 2: ditto

controllers/pkg/resource/meta.go
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright 2022 Nokia.
```
johnbelamaric, line 2: ditto

controllers/pkg/resource/meta_test.go
```diff
@@ -0,0 +1,52 @@
+/*
+Copyright 2022 Nokia.
```
johnbelamaric, line 2: ditto

johnbelamaric (commented):
Nice, almost there, but copyright needs to be fixed.

controllers/pkg/resource/annotations_test.go
henderiw: fixed

controllers/pkg/resource/finalizer.go
henderiw: fixed

controllers/pkg/resource/finalizer_test.go
henderiw: fixed

controllers/pkg/resource/meta_test.go
henderiw: fixed

PR #201, opened 2023-05-17
New approach for the central Makefile used by PROW

kispaljr:
PROW's central `Makefile` is changed to delegate commands (`make` targets) to the `Makefiles` of individual go modules.
This approach requires a `Makefile` to be present in the directory of all `go.mod` files.
This approach provides sensible defaults for the `make` targets that PROW uses, and even for other common tasks, but also allows each go module to provide its own unittest/lint/gosec/... implementation if necessary.

This PR also re-enables `gosec` presubmit tests, since it fixes the issue that was the reason for disabling them.


Makefile
```diff
@@ -14,42 +26,18 @@ CONTAINER_RUNTIME=docker
...
+# NOTE: the above line is complicated due to the limited capabilities of busybox's `find`.
+# It meant to be equivalent with this:  find . -name 'go.mod' -printf "'%h' " 
+
+
+.PHONY: unit lint gosec unit_clean test
+# delegate these commands to the Makefiles next to the go.mod files
+unit lint gosec unit_clean test: 
+	for dir in $(GO_MOD_DIRS); do \
+		$(MAKE) -C "$$dir" $@ ; \
+	done
```
electrocucaracha, line 42: Shouldn't be better to include the find instruction here? The only advantage could be to avoid some executions in other targets.
```suggestion
	for dir in $(shell find . -name 'go.mod' -exec sh -c 'echo \"$$(dirname "{}")\" ' \;); do \
		$(MAKE) -C "$$dir" $@ ; \
	done
```

controllers/specializer-operator/Makefile
```diff
@@ -16,15 +16,24 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+
+# Only enable bash if it is in the PATH: 
+# this check is needed to be compatible with the busybox container used by PROW presubmit tests (that doesn't contain bash)
+HAVE_BASH ?= $(shell bash --version > /dev/null 2>&1; echo $$?)
```
electrocucaracha, line 22: Another option could be to use `command -v`
```suggestion
HAVE_BASH ?= $(shell command -v bash > /dev/null 2>&1; echo $$?)
```

default-docker.mk
```diff
@@ -0,0 +1,21 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+.PHONY: docker-build
+docker-build:  ## Build docker image
+	docker buildx build --load --tag  ${IMG} -f ./Dockerfile .
+
+.PHONY: docker-push
+docker-push: ## Build and push docker image
+	docker buildx build --push --tag  ${IMG} -f ./Dockerfile .
```
electrocucaracha, line 21: What about declaring a common buildx variable and using it with different args?

electrocucaracha (commented):
Hi @kispaljr great job on this. I put some suggestions about it, but I noticed that some files contains an end file characters, maybe added by the editor. I'm not sure if those can affect the execution.

Makefile
kispaljr: I find it more readable as it is. Wouldn't you agree?

controllers/specializer-operator/Makefile
kispaljr: I accepted your change, will push it soon

kispaljr:
@electrocucaracha  Thank you for the review. I've tried to address your comments in the latest commit

default-docker.mk
```diff
@@ -0,0 +1,31 @@
...
+CONTAINER_RUNTIME=podman
+else
+CONTAINER_RUNTIME=docker
+endif
+
+endif
+
+.PHONY: docker-build
+docker-build:  ## Build docker image
+	$(CONTAINER_RUNTIME) build --load --tag  ${IMG} -f ./Dockerfile .
```
electrocucaracha, line 27: I got an error with the `--load` flag. Does this require a specific Docker version?
```
docker build --load --tag  docker.io/nephio/dnn-fn:latest -f ./Dockerfile .                                                                                                                                                                                                                                                
unknown flag: --load                                                                                                                                                                                                                                                                                                       
See 'docker build --help'. 
```

electrocucaracha (changes requested):

electrocucaracha:
\lgtm

gvbalaji:
Thanks @kispaljr .
/approve

gvbalaji:
@johnbelamaric @radoslawc @henderiw PTAL and approve to merge this.

PR #202, opened 2023-05-17
controller-manager framework with repo-controller example

ssharmasandeep:


henderiw:
Sandeep we should put reconcilers under 
controllers/pkg/reconcilers 
this avoids another go mod file.

Also to get this PR through the test chain we have to split it up.
First PR should be limited to the reconciler-interface. otherwise test will never succeed

ssharmasandeep:
@henderiw if I move the reconciler under pkg/ then the directory structure will be
pkg/
  go.mog
  reconciler-interface/
  reconciler1/
  reconcilern/

All packages part of same module i.e pkg.  Also the reconciler* will import the reconciler-interface package.. is this how you are suggesting the framework to be? 



what we discussed in the ticket was something like this,

pkg/
   reconciler-interface/
   go.mod
   reconciler1/
   go.mod
   reconcilern/
   go.mod

gvbalaji:
Rebase needed and  signing of CLA

PR #208, opened 2023-05-18
nephio-controller-manager operator

ssharmasandeep:
This PR has the operator nephio-controller-manager.  
Since we have to create  PR  one go.mod change at a time, the subsequent PRs post this will have the controllers registered
with the main operator manager.


controllers/pkg/reconcilers/config/config.go
```diff
@@ -19,12 +19,15 @@ package ctrlrconfig
 import (
 	"time"
 
+	"github.com/henderiw-nephio/nephio-controllers/pkg/giteaclient"
```
johnbelamaric, line 22: What's this package, should it be in a nephio repo?

controllers/pkg/reconcilers/config/config.go
```diff
@@ -19,12 +19,15 @@ package ctrlrconfig
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 )
 
 type ControllerConfig struct {
 	PorchClient client.Client
+	GiteaClient giteaclient.GiteaClient
```
johnbelamaric, line 29: Is this config object going to contain some sort of entries for each reconciler? We may want to refactor that at some point. But this is OK for now.

johnbelamaric (approved):
Looks good, a few non-blocking points.

/approve

controllers/pkg/reconcilers/config/config.go
ssharmasandeep: Yes . Changed tjhis to "github.com/nephio-project/nephio/controllers/pkg/giteaclient".

controllers/pkg/reconcilers/config/config.go
ssharmasandeep: Not for each reconciler, it will be meant for the repo controller.  
 I will refactor this when integrating repo-controller with controller-manager,  I am not sure if we expect this as env variable for that controller .

johnbelamaric:
Ok, thanks. That looks good but I would like another set of eyes on it.

gvbalaji:
Thanks @ssharmasandeep . Overall looks good.  Can you please add a unit test to have some coverage?

ssharmasandeep:
Bala, this is not complete functionality. I will add unit tests in
subsequent PRs in which we start registering controllers with this operator.
-Sandeep

ssharmasandeep:
/test presubmit-nephio-go-test



<br class="Apple-interchange-newline">

PR #212, opened 2023-05-22
reworked bootstrap controllers

henderiw:
I had to rework the bootstrap controllers and split them in 2. bootstrap packages and bootstrap secrets
the approach to only rely on secret does not reliably work on packages:
-  a secret could be available but the package could be approved later in which case the package never get installed. So changed this to rely on packages for the bootstrap controller.

updated readme(s) to reflect the changes.


controllers/pkg/reconcilers/bootstrap-packages/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## implementation
+
+The controller acts on package revision resources. It first figures out if the resources of a package revision are to be installed on the remote cluster, by checking if:
+- repository has the  `nephio.org/staging` key set
+
+If the controller knows the package is to be installed on the remote cluster it finds the cluster name by checking the `nephio.org/cluster-name` annotation of the first resource in the package. (we assume the `nephio.org/cluster-name` annotation is set on all resources). Once the controller knows the clustername it finds the credentials of the remote cluster and the type of cluster based on the signatures of the secret (right now only cluster api is implemented, but the code is able to handle other implementations).
+Once the remote credentials are found and the cluster is deemed ready, the package get installed on the remote cluster.
+
+If any of the validation fail the controller will retry installing the package. Right now the watch on package reviosions is a timed based loop.
```
liamfallon, line 13: reviosions ---> revisions

henderiw: fixed

liamfallon: /approve

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -209,45 +208,22 @@ func (r *reconciler) getResources(ctx context.Context, clusterName string) ([]un
...
 
-func MatchesStagingRepo(repoName string, stagingRepoNames []string) bool {
-	for _, stagingRepoName := range stagingRepoNames {
-		if repoName == stagingRepoName {
-			return true
-		}
+func (r *reconciler) getResources(ctx context.Context, req ctrl.Request) ([]unstructured.Unstructured, bool, error) {
+	prr := &porchv1alpha1.PackageRevisionResources{}
+	if err := r.porchClient.Get(ctx, req.NamespacedName, prr); err != nil {
+		r.l.Error(err, "cannot get package resvision resourcelist", "key", req.NamespacedName)
```
johnbelamaric, line 222: nit: s/resvision/revision/

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -88,36 +95,51 @@ type reconciler struct {
...
-		return reconcile.Result{}, nil
+		return ctrl.Result{}, nil
 	}
 
-	// if the secret is being deleted dont do anything for now
-	if cr.DeletionTimestamp != nil {
-		return reconcile.Result{}, nil
+	// check if the packagerevision is part of a staging repository
+	// if not we can ignore this package revision
+	stagingPR, err := r.IsStagingPackageRevision(ctx, cr)
```
johnbelamaric, line 112: I think we should check the package lifecycle to make sure it  is Published (there is an IsPublished method in PR I believe, as "Deletion Proposed" is also considered published.

johnbelamaric (commented):
Only think I see that needs fixing is to make sure the package is published.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: fixed

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: fixed

PR #214, opened 2023-05-22
KRM function SDK: separate generate and update callbacks

kispaljr:


henderiw:
We should not be doing this. We need to do 2 enhancements:
1. is a more advanced diff to allow the info that is available upstream to trickle down and not having to do another lookup of the same resource we already processed
2. the final object should be present in the package and the right owner reference should be set.

This allows to simplify the SDK rather than adding complexity. With 1/2 the code would no longer be needed at all.

kispaljr:
this was already fixed elsewhere

PR #226, opened 2023-05-28
Auto-approval controller

johnbelamaric:


johnbelamaric:
A few things about this PR:

- I don't know how the roles were generated, `make generate` fails. I do need some updated permissions, in particular for the approval subresource.
- Right now it does a replace for the pkg module. Otherwise I have to commit that before I can use. Maybe go workspaces can fix this. But for now, after this merges, I can do a follow up PR that sets the modules up right. Something is wrong with how we're doing this.
- I added a way to build a docker image locally (not in the Dockerfile) so that we can test when using `replace` - this will tag the image your user name and you can put it with that name. Works for now, a bit awkward.

johnbelamaric:
With this and https://github.com/nephio-project/nephio-example-packages/pull/35, we will be able to just apply a PVS like this:

```
apiVersion: config.porch.kpt.dev/v1alpha2
kind: PackageVariantSet
metadata:
  name: edge-clusters
spec:
  upstream:
    repo: nephio-example-packages
    package: nephio-workload-cluster
    revision: v3
  targets:
  - repositories:
    - name: mgmt
      packageNames:
      - edge1
      - edge2
      - edge3
      - edge4
    template:
      annotations:
        approval.nephio.org/policy: initial
```

and those 4 clusters will be automatically created, registered with Nephio, and bootstrapped (well, with #232 also)

PR #228, opened 2023-05-28
added controller runtime 15.0

henderiw:
updated related to controller runtime 15. As it has some differences


johnbelamaric:
Looks like it breaks some stuff :(

henderiw:
yes I am dealing with it. it has to do with the multiple go.mod files. So trying to figure out how to get around it.
I wanted to get this done asap, since there are indeed a lot of dependencies with this and the sooner we do it the better it will be for us I believe

henderiw:
@johnbelamaric @gvbalaji would like to merge this asap. as it touches the complete repo

kispaljr:
I agree that this should be merged soon.
Just one small cosmetic thing: could you remove the commented out "replace" lines from multiple `go.mod` files?

henderiw:
I removed the commented elements in the go.mod files

PR #253, opened 2023-06-02
Remove replace in go.mod

johnbelamaric:
In order for the automated docker build to work, we need to comment out the replace line.

Merging this will trigger the docker build and push the image to our registry.

Alternatively, instead of merging this we can any of these:
- Manually build and push
- Fix the docker build in the nephio-controller-manager to copy in the entire controllers directory and tweak the replace statement
- Figure out something better than the way we do our multiple go.mod files right now

### DIFF
--- a/operators/nephio-controller-manager/go.mod
+++ b/operators/nephio-controller-manager/go.mod
@@ -2,7 +2,7 @@ module github.com/nephio-project/nephio/operators/nephio-controller-manager
 
 go 1.20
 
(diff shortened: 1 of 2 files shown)


johnbelamaric:
The point is, any PR that merges that touches the controllers/pkg (like #252 for example) needs a sort of double-commit like this, once to get the pkg directory into github, then once to consume that from github rather than from the replace. Which is a pain.

PR #262, opened 2023-06-06
updated replace in go mod files/krm functions

henderiw:
updated replace in arm fn go mod file now that the lib is merged into main.


krm-functions/dnn-fn/Dockerfile
```diff
@@ -15,10 +15,8 @@
 FROM golang:1.20-alpine
 ENV CGO_ENABLED=0
 WORKDIR /go/src/
-COPY go.mod go.sum ./
-RUN go mod download
-COPY *.go ./
-COPY fn/*.go ./fn/
+COPY . .
```
radoslawc, line 18: Won't this: https://github.com/nephio-project/nephio/blob/a1e470f419a41873fd86895f4dce6ab1f4e9c804/krm-functions/dnn-fn/go.mod#L5

cause that 'lib' directory will have to also be present in container at /go/ directory?

henderiw: the go mod vendor does this. It is a fn of go mod to have local overrides.

PR #264, opened 2023-06-07
Preliminary check-in for NFTopology Controller

s3wong:


operators/nftopology/LICENSE
johnbelamaric, line 1: We probably don't need this as it already exists in the root of the repo.

operators/nftopology/Makefile
```diff
@@ -0,0 +1,157 @@
+
+# Image URL to use all building/pushing image targets
+IMG ?= controller:latest
```
johnbelamaric, line 3: we should name this nftopology-controller or something.

operators/nftopology/PROJECT
```diff
@@ -0,0 +1,16 @@
+# Code generated by tool. DO NOT EDIT.
+# This file is used to track the info used to scaffold your project
+# and allow the plugins properly work.
+# More info: https://book.kubebuilder.io/reference/project-config.html
+domain: nephio.org
+layout:
+- go.kubebuilder.io/v3
+projectName: nftopology
+repo: nephio-project/nftopology
```
johnbelamaric, line 9: Maybe we should keep it int hat repo? Just for expediency, if there are build issues from moving it?

operators/nftopology/config/default/kustomization.yaml
```diff
@@ -0,0 +1,72 @@
+# Adds namespace to all resources.
+namespace: nftopology-system
```
johnbelamaric, line 2: Should run in nephio-system I think.

operators/nftopology/config/default/kustomization.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: nftopology-
```
johnbelamaric, line 9: I think this works to fix the manifests, but does it fix the Makefile variable? I am not sure

operators/nftopology/config/._default
johnbelamaric, line 1: There are tons of these binary files. I assume they are from your editor or something? Please remove them from the PR.

operators/nftopology/controllers/nftopology_controller.go
```diff
@@ -0,0 +1,182 @@
+/*
+Copyright 2023.
```
johnbelamaric, line 2: The Nephio Authors

operators/nftopology/controllers/nftopology_controller.go
```diff
@@ -0,0 +1,182 @@
...
+
+	pvv1alpha1 "github.com/GoogleContainerTools/kpt/porch/controllers/packagevariants/api/v1alpha1"
+	pvsv1alpha2 "github.com/GoogleContainerTools/kpt/porch/controllers/packagevariantsets/api/v1alpha2"
+	"github.com/go-logr/logr"
+	deployv1alpha1 "github.com/nephio-project/api/nf_deployments/v1alpha1"
+	reqv1alpha1 "github.com/nephio-project/api/nf_requirements/v1alpha1"
+)
+
+const (
+	FREE5GC            string = "free5gc"
```
johnbelamaric, line 38: I think in Go it is usual to use CamelCase for constants.

operators/nftopology/controllers/nftopology_controller.go
```diff
@@ -0,0 +1,182 @@
...
+	} else {
+		return false
+	}
+}
+
+/*
+ * BuildPVS: builds a PVS object from NFTopology's NFInstance
+ * NFClass needed for package reference
+ */
+func BuildPVS(nfInst *reqv1alpha1.NFInstance, nfClassObj *reqv1alpha1.NFClass) *pvsv1alpha2.PackageVariantSet {
```
johnbelamaric, line 86: Is there a reason these (AllUPFsDeployed, BuildPVS, etc.) are exported (capitalized)? Are they called by another module? If not, they should be local (lowercase first letter).

operators/nftopology/controllers/nftopology_controller.go
```diff
@@ -0,0 +1,182 @@
...
+
+	retPVS.Spec.Targets = append(retPVS.Spec.Targets, *target)
+
+	// TODO(s3wong): adding Template in PVS?
+	return retPVS
+}
+
+//+kubebuilder:rbac:groups=req.nephio.org,resources=nftopologies,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=req.nephio.org,resources=nftopologies/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=req.nephio.org,resources=nftopologies/finalizers,verbs=update
```
johnbelamaric, line 118: I think you need some CRUD privileges on PVS as well.

operators/nftopology/controllers/nftopology_controller.go
```diff
@@ -0,0 +1,182 @@
...
+func (r *NFTopologyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.l = log.FromContext(ctx).WithValues("req", req)
+	r.l.Info("Reconcile NFTopology")
+
+	nfTopo := &reqv1alpha1.NFTopology{}
+	if err := r.Client.Get(ctx, req.NamespacedName, nfTopo); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	nfDeployed := &deployv1alpha1.NFDeployed{}
```
johnbelamaric, line 131: I'm still a little confused as to what NFDeployed is. It almost seems like it should be part of NFTopology.Status. Or is the idea to have many of them for one NFTopology, in which case I can see using a separate resource.

operators/nftopology/controllers/nftopology_controller.go
```diff
@@ -0,0 +1,182 @@
...
+	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: nfTopo.ObjectMeta.Namespace, Name: nfTopo.ObjectMeta.Name}, nfDeployed); err != nil {
+		r.l.Info(fmt.Sprintf("NF Deployed object for %s not created, continue...\n", nfTopo.ObjectMeta.Name))
```
johnbelamaric, line 133: Should you be returning here? Or is it OK to continue with the empty one?

operators/nftopology/controllers/nftopology_controller.go
```diff
@@ -0,0 +1,182 @@
...
+				// skip this SMF
+				continue
+			}
+		}
+
+		pvs := BuildPVS(&nfInst, nfClass)
+		r.l.Info(fmt.Sprintf("PVS for NF inst %s is %+v\n\n\n", nfInst.Name, pvs))
+		if err := r.Client.Create(ctx, pvs); err != nil {
+			r.l.Error(err, fmt.Sprintf("Failed to create PVS %s: %s\n", nfInst.Name, err.Error()))
+		} else {
```
johnbelamaric, line 161: I'm confused by this else - isn't this saying: if I have an error during create, log it; otherwise update the thing I just created.

operators/nftopology/controllers/packagerevision_controller.go
```diff
@@ -0,0 +1,258 @@
...
+	l := log.FromContext(ctx).WithValues("NFTopology-PackageRevision", req.NamespacedName)
+
+	pkgRev := &porchv1alpha1.PackageRevision{}
+	if err := r.Get(ctx, req.NamespacedName, pkgRev); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Step 1: check package revision state
+	// TODO(s3wong): if package deleted, inform ESA
+	if pkgRev.Spec.Lifecycle != porchv1alpha1.PackageRevisionLifecyclePublished {
```
johnbelamaric, line 166: You should use `porchv1alpha1.LifecycleIsPublished(pkgRev.Spec.Lifecycle)`, as `DeletionProposed` is also still considered "Published".

operators/nftopology/controllers/packagerevision_controller.go
```diff
@@ -0,0 +1,258 @@
...
+		l.Info("Package lifecycle is not Published, ignore", "PackageRevision.lifecycle",
+			pkgRev.Spec.Lifecycle)
+		return ctrl.Result{}, nil
+	}
+
+	var regionName, clusterName, nfInstanceName, topologyName, topoNamespace string
+	var createDeployed, found bool
+	pkgRevLabels := pkgRev.ObjectMeta.Labels
+	if regionName, found = pkgRevLabels["region"]; !found {
```
johnbelamaric, line 175: `nephio.org/region`, though I don't think we set this now

operators/nftopology/controllers/packagerevision_controller.go
```diff
@@ -0,0 +1,258 @@
...
+		l.Info("region name not found from package")
+		return ctrl.Result{}, nil
+	}
+	if nfInstanceName, found = pkgRevLabels["nfinstance"]; !found {
```
johnbelamaric, line 179: If these are ones you create, you should namespace them. Probably `nftopology.nephio.org/nfinstance`, etc.

operators/nftopology/controllers/packagerevision_controller.go
```diff
@@ -0,0 +1,258 @@
...
+	if topologyName, found = pkgRevLabels["nftopology"]; !found {
+		l.Info("NF Topology name not found from package")
+		return ctrl.Result{}, nil
+	}
+	if topoNamespace, found = pkgRevLabels["nftopology-namespace"]; !found {
+		l.Info("NF Topology namespace not found from package")
+		return ctrl.Result{}, nil
+	}
+	// repo name is the cluster name
+	clusterName = pkgRev.Spec.RepositoryName
```
johnbelamaric, line 192: That'll work for R1 but we will break that 1:1 association at some point, most likely.

operators/nftopology/controllers/packagerevision_controller.go
```diff
@@ -0,0 +1,258 @@
...
+		err := errors.New(fmt.Sprintf("%s not found in NF Topology %s\n", nfInstanceName, topologyName))
+		l.Error(err, fmt.Sprintf("NF Instance %s not found in NF Topology %s\n", nfInstanceName, topologyName))
+		return ctrl.Result{}, err
+	}
+	// Get NFClass, cluster scope
+	className := nfInstance.NFTemplate.ClassName
+	var nfClass = &nfreqv1alpha1.NFClass{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: className}, nfClass); err != nil {
+		l.Error(err, fmt.Sprintf("NFClass object not found: %s: %s\n", className, err.Error()))
+		return ctrl.Result{}, err
```
johnbelamaric, line 230: I think this is fine for now, but eventually we may want to check things like this in a Validate method. What we do in a few other instances is we check references and stuff in Validate, and set the Stalled condition to True if there's dangling references.

operators/nftopology/controllers/packagerevision_controller.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	if err := BuildNFDeployed(deploymentName, topoNamespace, nfTopology, nfInstance, clusterName, nfClass.Spec.Vendor, nfClass.Spec.Version, nfDeployed); err != nil {
+		l.Error(err, fmt.Sprintf("Failed to build NFDeployed %s: %s\n", deploymentName, err.Error()))
+		return ctrl.Result{}, err
```
johnbelamaric, line 234: For many of these errors, we should instead be setting the status conditions (probably a Ready condition) and using the message there. Then, the user can see the errors via `kubectl`. If we just return them like this, they may only end up in the log, which is hard to debug.

So, if you look at something like the PackageVariant controller, you will see that right in the beginning of Reconcile, we setup a `defer` call that updates the Status. Then, throughout the method, we set the conditions in various ways. When we exit, those conditions get written to the status in the api server.

Another thing you can do, but we have not in most places, is send events. I do this in the approval controller (in the nephio/controllers/pkg/reconcilers dir), because I don't "own" the resources I am messing with and so chose not to add a condition. I could have added a condition, that would be legit, but instead I decided to just send events. Those show up when you do a "describe" on the primary resource using kubectl.

johnbelamaric (changes requested):
Great to see this PR! Here's a first review, I think there's a bit of cleanup but the basics are here. I haven't fully grokked it yet, I think it will help when I get a chance to try it out.

PR #269, opened 2023-06-08
Specializer updates

henderiw:
Updates in the specialised
- dealing with ipv4 and ipv6
- fixes bugs


krm-functions/dnn-fn/fn/function.go
```diff
@@ -190,8 +190,10 @@ func (f *dnnFn) updateDnnResource(dnnObj_ *fn.KubeObject, owned fn.KubeObjects)
 
 func getAnnotations(annotations map[string]string) map[string]string {
 	a := map[string]string{}
-	for k, v := range annotations {
-		a[k] = v
+	if len(annotations) > 0 {
```
johnbelamaric, line 193: this shouldn't be needed, range should handle nil and zero length just fine. Why was this necessary?

krm-functions/dnn-fn/fn/function.go
```diff
@@ -190,8 +190,10 @@ func (f *dnnFn) updateDnnResource(dnnObj_ *fn.KubeObject, owned fn.KubeObjects)
...
+		for k, v := range annotations {
+			a[k] = v
+		}
 	}
 	if owner, ok := annotations[condkptsdk.SpecializerPurpose]; ok {
```
johnbelamaric, line 198: This may blow up if annotations is nil, though (not 100% on that)

krm-functions/dnn-fn/fn/function.go
```diff
@@ -190,8 +190,10 @@ func (f *dnnFn) updateDnnResource(dnnObj_ *fn.KubeObject, owned fn.KubeObjects)
...
 		a[condkptsdk.SpecializerPurpose] = owner
```
johnbelamaric, line 199: didn't we just set this in the loop above?

krm-functions/interface-fn/fn/function.go
```diff
@@ -307,8 +307,10 @@ func (f *itfceFn) getAnnotationsWithvlanClaimName(itfce *nephioreqv1alpha1.Inter
 
 func getAnnotations(annotations map[string]string) map[string]string {
 	a := map[string]string{}
-	for k, v := range annotations {
-		a[k] = v
+	if len(annotations) > 0 {
```
johnbelamaric, line 310: same point...this seems weird

krm-functions/dnn-fn/fn/function.go
henderiw: I did this change and nothing worked any longer so was a bit paranoid and added this, but it did not change anything. So will revert

johnbelamaric: been there, done that :)

johnbelamaric (approved):
There's a lot here...I gave it a once over but I am relying a lot on the tests.

/approve
/lgtm

PR #278, opened 2023-06-11
connect-fn

henderiw:
connect-fn to provide the topology information how the cluster is connected.
Based on workload cluster name edge/region and number the fn generates a link and endpoint CR.
Link will be picked up by the wiring function to interconnect the kind node to the network
EP is used to generate the config of the network.

The fn generates the connectivity map based on a certain startegy. Here since we have a single network element
edge01 uses ports 1-4
edge02 uses port 5-7
edge03 uses port 8-11
edge04 uses port 12-15
edge is seen as 0 and uses port 1-4
-> we support max 4 edges since the conncectivity on the switch/router is limited
region01 uses ports 17-20
region02 uses port 21-24
region03 uses port 25-28
region04 uses port 29-32
region is seen as 0 and uses port 17-20

no we cannot have names different than edge or edgeXX or region and regionXX where XX is representing a number

We should make this configurable using a fn-config going forward.


johnbelamaric:
Ok, a couple comments/questions:

1) This is specific to our current demo topology and CAPI. Which is fine I think for now, at least we have something to wire the network in the demo. I think eventually this most likely the equivalent of this runs on the workload cluster as an operator so it can react to cluster autoscaling, for example. 
2) Does this require some new actuation controller running? If so, we'll need a package for that.
3) I see "clab" referenced, but does this use container lab or is it independent of that?

liamfallon:
This PR is rather large, it will take some time to review.

efiacor:
Hi @henderiw , is this still relevant?

PR #282, opened 2023-06-12
network reconciler

henderiw:
added network reconciler.
It creates the IPAM resources per network based on the network CRD
it also assigns VLANs if requested on the network.

The selector in the network CRD allows to select prefixes/subnets per cluster.
The network controller also generates a network config that can be picked up be a controller to actuate on the target network device.


controllers/pkg/reconcilers/network/reconciler.go
```diff
@@ -0,0 +1,341 @@
...
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/go-logr/logr"
+	configv1alpha1 "github.com/henderiw-nephio/network/apis/config/v1alpha1"
+	infra2v1alpha1 "github.com/henderiw-nephio/network/apis/infra2/v1alpha1"
+	"github.com/henderiw-nephio/network/pkg/endpoints"
```
johnbelamaric, line 27: Should these point to a nephio repo somewhere? Have those pieces been moved or is this part of the backend piece that hasn't moved yet?

johnbelamaric:
> No I can move these if you want. This is separate, but this creates a dependency. So let's test this. I will create a parallel PR for the api and after we update this repo.

Ok, I take that to mean "merge this PR". :)

/approve
/lgtm

PR #289, opened 2023-06-13
Nad default network

henderiw:
fixes for nad in default network.
Since we don't do the intelligent diff the NAD is presented with interfaces for which it does not have to do anything. This PR handles this case.
1. check in the NAD function -> default network
2. the sdk need to handle return of nil object from updateFn because of this.


johnbelamaric:
Nice - can you fix the linter error?

PR #291, opened 2023-06-13
Add a check for deltas before saving

johnbelamaric:
The auto-approval controller was not kicking in, because the PackageRevision creation timestamp kept updating. Which is odd, and probably a bug in Porch. But, we want this anyway. It will avoid saving the package revision resources if there is no delta. This should prevent the forever growing task list, hopefully.

Still testing - one concern is that if we are refreshing claims, we may be stuffing away a timestamp which means this would never find matching hashes.

### DIFF
--- a/controllers/pkg/porch/util/packagerevision.go
+++ b/controllers/pkg/porch/util/packagerevision.go
@@ -1,0 +1,35 @@
+/*
+ Copyright 2023 The Nephio Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package util
+
+import (
+	"context"
+	"crypto/sha1"
+	"encoding/hex"
+
+	porchv1alpha1 "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+func PackageRevisionResourcesHash(prr *porchv1alpha1.PackageRevisionResources) (string, error) {
+	b, err := yaml.Marshal(prr.Spec.Resources)
+	if err != nil {
+		return "", err
+	}
+	hash := sha1.Sum(b)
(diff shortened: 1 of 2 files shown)


johnbelamaric:
This is probably still useful to have, but it is not fixing the problem yet.

What I have found is that when the package variant controller saves the Kptfile to package revision resources, the Kptfile keys (name, description, etc.) and the condition keys (type, message, etc.) are in one order, while when the specializer does it they are in a different order. Both do sort the condition arrays themselves; the issue is the field names come out in different orders.

Suspiciously, they use different versions of kyaml (13.9 in porch, 14.2 in nephio), which is what serializes that. Though I didn't immediately spot differences that would cause this.

I am trying to see if I can build a version of porch-controllers with the newer kyaml, and also with some additional instrumentation. I think there are some fixes to the way existing revisions are processed that may also contribute.

johnbelamaric:
Ugh. Building porch-controllers with an updated kyaml is a nightmare, it descends into updating k8s API, kustomize api, controller-runtime, etc.

I will just add my instrumentation.

Right now, what this means is that we have to manually approve the NF packages and other packages that go through specialization. Auto-approve won't do it. But we can work around that.

johnbelamaric:
This is the method that seems to re-order the conditions:

https://github.com/GoogleContainerTools/kpt/blob/main/porch/controllers/packagevariants/pkg/controllers/packagevariant/injection.go#L284

Not 100% sure but checking now.

henderiw:
I will have a look. We did add something to sort conditions since the tests validations need something otherwise they fail as we check the expected result.

henderiw:
Here is where we sort conditions.

// SortConditions returns a kptfile with sorted conditions
func (r *kptFile) SortConditions() {
	r.m.RLock()
	defer r.m.RUnlock()

	if r.GetKptFile().Status == nil || len(r.GetKptFile().Status.Conditions) == 0 {
		return
	}

	sort.SliceStable(r.GetKptFile().Status.Conditions, func(i, j int) bool {
		return r.GetKptFile().Status.Conditions[i].Type < r.GetKptFile().Status.Conditions[j].Type
	})
}

johnbelamaric:
/hold

Still testing this, but may not be needed given #295 - may still also be helpful, but if not needed we may want to defer it.

efiacor:
Hi @johnbelamaric , is this still relevant/needed?

liamfallon:
What is the status of this? We should check if it's still an issue and fix in R4.

efiacor:
@kispaljr looks like we still have differences in the kptfile rendering. Did the sdk fix not address this?
#295

liamfallon:
The bug regarding the PackageRevision timestamp constantly updating mentioned by @johnbelamaric above has been fixed in Porch, see https://github.com/nephio-project/porch/pull/180

liamfallon:
Porch has been moved to kptdev/porch.

PR #295, opened 2023-06-14
Fix SDK so that it keeps formatting of Kptfile

kispaljr:


krm-functions/interface-fn/fn/testdata/interface_cni_vlan-ipv6only/_expected.yaml
```diff
@@ -32,16 +32,25 @@ items:
...
+    #commentB
     annotations:
       config.kubernetes.io/local-config: "true"
-    name: pkg-upf
+  info:
+    description: upf package example
   pipeline: {}
   status:
     conditions:
+    - message: update condition for initial resource
```
johnbelamaric, line 47: I am not 100% sure this PR is going to fix the probelm we are seeing. The problem I see is that when the Porch controllers emit the Kptfile, these come out like:

```
- type: workload.nephio.org/v1alpha1.UPFDeployment.upf-cluster01
  status: "False"
  message: update for condition
```

whereas here they are: message, status, type.

That said, maybe just not changing the ordering will fix it? I am not sure.

kispaljr: hmmm... I see
right now we keep the order of fields of the conditions that already exist in the input's Kptfile, but for the new conditions added to the list by the KRM functions the order of fields are random/arbitrary

johnbelamaric: Example of commit by *nephio-controller*:

image


Example of follow up commit "fixing" the order, by *package variant controller*:

image

Notice how not only the fields of the conditions re-order, but even the Kptfile `description` moves.

If this PR keeps Kptfile ordering in this sense (field-by-field), then it should help.

kispaljr: where should be the "reason" field in the above order? (type, status, message)

johnbelamaric: If we keep the existing ordering, then I think we will be ok. It may thrash once or twice, but then settle down. That is good enough for now.

kispaljr: but if the problem was that KRM functions are "fighting" with Porch, where both modifies the order of fields back-and-forth: this shouldn't happen anymore. 
What should happen instead is that Porch modifies the field order of the condition created by a KRM function, but after  the KRM functions should leave the field order as-is

johnbelamaric: > right now we keep the order of fields of the conditions that already exist in the input's Kptfile, but for the new conditions added to the list by the KRM functions the order of fields are random/arbitrary

The last comment is in response to this. If this is true, of the condition fields and the Kptfile fields (e.g, description), then we should be OK.

kispaljr: that is the theory, yes ...

let's see it in practice :)

johnbelamaric: ok!

/approve
/lgtm

PR #296, opened 2023-06-14
Create NADs with a namespace

johnbelamaric:
Add a default namespace to NADs when creating the, so set-namespace will act upon them.


johnbelamaric:
@kispaljr @henderiw @tliron FYI

johnbelamaric:
Merging this.

PR #300, opened 2023-06-14
Improve kpt package (Kptfile) condition handling

kispaljr:
It also sorts the conditions in the Kptfile's status by type

### DIFF
--- a/krm-functions/dnn-fn/fn/testdata/create_owned_resources/_expected.yaml
+++ b/krm-functions/dnn-fn/fn/testdata/create_owned_resources/_expected.yaml
@@ -46,13 +46,13 @@ items:
   pipeline: {}
   status:
     conditions:
-    - message: update for condition
+    - type: req.nephio.org/v1alpha1.DataNetwork.internet
       status: "False"
-      type: req.nephio.org/v1alpha1.DataNetwork.internet
-    - message: create resource
+      message: update for condition
+    - type: ipam.resource.nephio.org/v1alpha1.IPClaim.internet-pool1
+      status: "False"
+      message: create resource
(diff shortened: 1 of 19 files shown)


gvbalaji (commented):
/approve

johnbelamaric:
@kispaljr this looks like it will fix the problem I worked around with https://github.com/GoogleContainerTools/kpt/pull/3990/files#diff-d81df890ed5e08ca074d2063624c3fbb25bfb1033030685dcd40799146e4356eR808, since it changes the output of those fields. I did also see the Kptfile "description" orderered differently between the Nephio and package variant controllers too, so I am not sure it will solve it completely.

I think we may want to hold off on this until after R1, unless we believe it is necessary for some other treason.

kispaljr:
OK, let's not break things and wait until R1.
With this I intended to remove every piece of code from the SDK that rewrites the Kptfile resource. The only way I update the Kptfile is by directly modifying the list items under `status.conditions`, so in theory it shouldn't even touch anything outside that. But to be honest this should have been true for my previous attempt, as well, so if you've seen the description field changed by a KRM function than it probably means there is bug still in the code.

efiacor:
Hi @kispaljr , is this change still relevant?

kispaljr:
> Hi @kispaljr , is this change still relevant?

well, in order to answer that, I have to investigate if the problem that it tried to solve still exist or not. 
We had a workaround on porch's side, but this would improve things on the kpt functions' side

kispaljr:
Yes, this is still relevant, but the library itself changed considerably underneath.
I will rebase the code to the new library.

PR #312, opened 2023-06-16
Specialization name namespace

henderiw:


kispaljr:
Instead of adding the "for" annotation to the names of owned objects, wouldn't it be better to simply add the kind and namespace of the owner object? 
That information is always available, while I am not sure if the "for" annotation have a meaning in all use cases. I understand it in the case of NFDeployment kpt packages, but these KRM functions might be used in other contexts, as well, where "for" doesn't have a meaningful value.

PR #321, opened 2023-06-27
updated configinject to only take into account published packages

henderiw:
updated ocnfiginject to only take into account published packages

### DIFF
--- a/controllers/pkg/reconcilers/generic-specializer/reconciler.go
+++ b/controllers/pkg/reconcilers/generic-specializer/reconciler.go
@@ -292,10 +292,14 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		kptf := kptfilelibv1.KptFile{Kptfile: rl.Items.GetRootKptfile()}
 		pr.Status.Conditions = porchcondition.GetPorchConditions(kptf.GetConditions())
-		if err = r.porchClient.Status().Update(ctx, pr); err != nil {
+		if err = r.porchClient.Update(ctx, pr); err != nil {
+			r.recorder.Event(pr, corev1.EventTypeWarning, "ReconcileError", "cannot update pr status")
+			r.l.Error(fmt.Errorf("cannot update pr status"), "")
 			return ctrl.Result{}, err
 		}
 		if err = r.porchClient.Update(ctx, prr); err != nil {
+			r.recorder.Event(pr, corev1.EventTypeWarning, "ReconcileError", "cannot update packagerevision resources")
+			r.l.Error(fmt.Errorf("cannot update packagerevision resources"), "")
 			return ctrl.Result{}, err
 		}
 	}
--- a/controllers/pkg/reconcilers/generic-specializer/reconciler.go
+++ b/controllers/pkg/reconcilers/generic-specializer/reconciler.go
@@ -286,7 +286,7 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		kptfile := rl.Items.GetRootKptfile()
 		if kptfile == nil {
 			r.recorder.Event(pr, corev1.EventTypeWarning, "ReconcileError", "mandatory Kptfile is missing")
-			r.l.Error(fmt.Errorf("mandatory Kptfile is missing from the package"), "")
+			r.l.Error(err, "mandatory Kptfile is missing from the package")
 			return ctrl.Result{}, nil
 		}
 
@@ -294,12 +294,12 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		pr.Status.Conditions = porchcondition.GetPorchConditions(kptf.GetConditions())
 		if err = r.porchClient.Update(ctx, pr); err != nil {
 			r.recorder.Event(pr, corev1.EventTypeWarning, "ReconcileError", "cannot update pr status")
-			r.l.Error(fmt.Errorf("cannot update pr status"), "")
+			r.l.Error(err, "cannot update pr status")
 			return ctrl.Result{}, err
 		}
 		if err = r.porchClient.Update(ctx, prr); err != nil {
 			r.recorder.Event(pr, corev1.EventTypeWarning, "ReconcileError", "cannot update packagerevision resources")
-			r.l.Error(fmt.Errorf("cannot update packagerevision resources"), "")
+			r.l.Error(err, "cannot update packagerevision resources")
 			return ctrl.Result{}, err
 		}
 	}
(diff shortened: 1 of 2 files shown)


krm-functions/configinject-fn/fn/function.go
```diff
@@ -139,84 +140,118 @@ func (f *FnR) desiredOwnedResourceList(forObj *fn.KubeObject) (fn.KubeObjects, e
...
+				// only update if the PR is published
+				if porchv1alpha1.LifecycleIsPublished(pr.Spec.Lifecycle) {
+					if !porchv1alpha1.LifecycleIsPublished(latestPR.Spec.Lifecycle) {
+						// if latest entry in the prmap is not publish but the new pr is published we update the map
+						prmap[prName] = pr
+					} else {
+						// both the latest pr and the new pr are published
+						// update the map with the latest pr
+						// if the revision of the new pr is better than the one of the latest pr in the map
+						if pr.Spec.Revision > latestPR.Spec.Revision {
```
johnbelamaric, line 177: "Revision" is a string, and right now anyway it will be "vN" where N is an integer. So, this will result in bogus ordering if there are more than 9 revisions.

krm-functions/configinject-fn/fn/function.go
```diff
@@ -139,84 +140,118 @@ func (f *FnR) desiredOwnedResourceList(forObj *fn.KubeObject) (fn.KubeObjects, e
...
 
 	resources := fn.KubeObjects{}
-	// walk through all the package revisions and check if the dependent resources are ready
+	// walk through all the package revisions and build a map of the pr(s) that
+	// have the packagename and are in a repo that has deployment true
+	// The map will contain the latest published revision of the pr, if no pr
+	// is published it will have a reference to this pr
 	// we assume there needs to be 1 dependency that resolves
-	found := false
+	prmap := map[string]porchv1alpha1.PackageRevision{}
```
johnbelamaric, line 148: Should this be a map of pointers and we do DeepCopy, instead of a map of structs?

controllers/pkg/reconcilers/generic-specializer/reconciler.go
```diff
@@ -245,24 +245,26 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-							filename = fmt.Sprintf("%s/%s", o.GetNamespace(), filename)
+							filename = fmt.Sprintf("%s/%s.yaml", o.GetNamespace(), filename)
 						}
 						prr.Spec.Resources[filename] = o.String()
 					} else {
 						prr.Spec.Resources[o.GetAnnotation(kioutil.PathAnnotation)] = o.String()
 					}
 					r.l.Info("generic specializer", "kind", own.Kind, "pathAnnotation", o.GetAnnotation(kioutil.PathAnnotation))
-
 					r.l.Info("generic specializer config injector", "cluserName", clusterName, "resourceName", fmt.Sprintf("%s/%s", own.Kind, o.GetName()))
```
johnbelamaric, line 257: s/cluserName/clusterName/

controllers/pkg/reconcilers/generic-specializer/reconciler.go
```diff
@@ -286,13 +288,24 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-			r.l.Error(fmt.Errorf("mandatory Kptfile is missing from the package"), "")
+			r.l.Error(err, "mandatory Kptfile is missing from the package")
 			return ctrl.Result{}, nil
 		}
 
+		r.l.Info("generic specializer root kptfile", "packageName", pr.Spec.PackageName, "repository", pr.Spec.RepositoryName, "kptfile", kptfile)
+
 		kptf := kptfilelibv1.KptFile{Kptfile: rl.Items.GetRootKptfile()}
 		pr.Status.Conditions = porchcondition.GetPorchConditions(kptf.GetConditions())
+		/*
```
johnbelamaric, line 299: If we don't need to update the PR at all, we can clean out some of this code. Fine perhaps for now to just put a TODO in with an explanation.

krm-functions/configinject-fn/fn/function.go
```diff
@@ -139,84 +140,118 @@ func (f *FnR) desiredOwnedResourceList(forObj *fn.KubeObject) (fn.KubeObjects, e
...
 
-				depObjs := rl.Items.Where(fn.IsGroupVersionKind(ref.GroupVersionKind()))
-				if len(depObjs) == 0 {
-					fn.Logf("configinject dependency not ready: the package %s in repo %s does not contain a resource with %s\n", pr.Spec.PackageName, pr.Spec.RepositoryName, gvk.String())
-					return nil, fmt.Errorf("dependency not ready: the package %s in repo %s does not contain a resource with %s", pr.Spec.PackageName, pr.Spec.RepositoryName, gvk.String())
+		// get the dependency objects in the package and check its status
+		gvk := schema.GroupVersionKind{
+			Group:   nephiodeployv1alpha1.GroupVersion.Group,
+			Version: nephiodeployv1alpha1.GroupVersion.Version,
+			Kind:    nephiodeployv1alpha1.UPFDeploymentKind,
```
johnbelamaric, line 218: Eventually this will need to be generic, but that's R1++ or R2 I guess?

johnbelamaric (changes requested):

krm-functions/configinject-fn/fn/function.go
henderiw: ok fixed it

krm-functions/configinject-fn/fn/function.go
henderiw: yes I need to think how to do this.

krm-functions/configinject-fn/fn/function.go
```diff
@@ -139,84 +141,120 @@ func (f *FnR) desiredOwnedResourceList(forObj *fn.KubeObject) (fn.KubeObjects, e
...
+					// both the latest pr and the new pr are published
+					// update the map with the latest pr
+					// if the revision of the new pr is better than the one of the latest pr in the map
+					newRev, err := getRevisionNbr(pr.Spec.Revision)
+					if err != nil {
+						return nil, err
 					}
-					// encapsulates the resource in another CR
-					newObj, err := GetConfigKubeObject(forObj, o)
+					latestRev, err := getRevisionNbr(latestPR.Spec.Revision)
```
johnbelamaric, line 173: Is this going to blow up for the "main" revision? Or for Drafts? should we map those to revision number 0 or something?

johnbelamaric: Actually I just verified, the latest revision will have an annotation:

```
  labels:
    kpt.dev/latest-revision: "true"
```

Of course...since updates to multiple objects are non-atomic, maybe relying on your comparison is better (ie, the API server cannot guarantee that exactly one package revision has that annotation at any given time).

PR #347, opened 2023-08-18
Basic config map generator function

johnbelamaric:


henderiw:
could we also add negative tests? like no key provided.

johnbelamaric:
Example of usage: https://github.com/nephio-project/nephio-packages/pull/31

vjayaramrh:
Any instructions on what steps can be tested out on a laptop or a container? Thanks

johnbelamaric:
> Any instructions on what steps can be tested out on a laptop or a container? Thanks

There is an image built and published to the Nephio Docker Hub. This PR makes use of it: https://github.com/nephio-project/nephio-packages/pull/31

You could copy what is done in there or in the unit tests in this PR, and execute the function using `kpt`.

krm-functions/gen-configmap-fn/fn/gen_configmap.go
```diff
@@ -0,0 +1,136 @@
...
+		name = rl.FunctionConfig.GetName()
+	}
+
+	err = fc.Validate()
+	if err != nil {
+		return false, err
+	}
+
+	cmko := fn.NewEmptyKubeObject()
+	err = cmko.SetAPIVersion("v1")
```
efiacor, line 91: Is the version likely to change any time soon?

efiacor (commented):
Minor comments

krm-functions/gen-configmap-fn/fn/gen_configmap.go
```diff
@@ -0,0 +1,136 @@
...
+limitations under the License.
+*/
+
+package fn
+
+import (
+	"bytes"
+	"fmt"
+	"text/template"
+	//"github.com/google/safetext/yamltemplate"
```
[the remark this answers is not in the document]
johnbelamaric: yeah, probably. I think I was planning to have a yaml rendering option that would use this library, which protects against yaml injection. but right now the function is ignorant of what format the generated text is in

krm-functions/gen-configmap-fn/fn/gen_configmap.go
johnbelamaric: No

johnbelamaric: (and even if a ConfigMap v2 came out, which would be several years from now, the v1 would still be valid)

krm-functions/gen-configmap-fn/fn/gen_configmap.go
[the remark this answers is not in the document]
efiacor: Yep, thought so. Could leave it as a bread crumb for later so.

PR #406, opened 2023-10-02
Fix for failing fossology scans

radoslawc:
Fix for https://github.com/nephio-project/sig-release/issues/43


radoslawc:
Closing current because of CLA error persisting here, I'll create new one.

PR #410, opened 2023-10-13
removing gitea related config from nephio manager module

vjayaramrh:


vjayaramrh:
@radoslawc not sure whey the fossology test is failing, any guidance on next steps is appreciated.

operators/nephio-controller-manager/main.go
```diff
@@ -149,16 +147,10 @@ func main() {
...
-	// of the manager of this controller cannot be used at this point.
-	// Should this be conditional ? Only if we have repo/token reconciler
-	g := giteaclient.New(resource.NewAPIPatchingApplicator(porchClient))
-	go g.Start(ctx)
 
 	ctrlCfg := &ctrlrconfig.ControllerConfig{
-		Address:         backendAddress,
-		GiteaClient:     g,
+		Address: backendAddress,
+		//GiteaClient:     g,
```
johnbelamaric, line 153: just go ahead and remove it rather than commenting it out.

johnbelamaric (commented):
Awesome, one little comment.

henderiw:
The issue with this approach right now is we have multiple clients from multiple controllers. We should avoid this. This is why I suggested to start the client the first one that comes

johnbelamaric:
So you would use a module level variable or something to share the client?

vjayaramrh:
> The issue with this approach right now is we have multiple clients from multiple controllers. We should avoid this. This is why I suggested to start the client the first one that comes

@henderiw I am guessing you are suggesting having a singleton that returns the same instance of client to every controller.
 Also, curious to know what you would consider the impact of each controller creating its own gitea client? Thanks

vjayaramrh:
> I was suggesting this to Vish. You have a generic library which you import in all the controllers. The First one that needs a client creates it and start the go routine. Afterwards if the 2nd controller comes it will reuse the one that is already available.
> 
> We can afterwards extend it to other git providers. As such we have 1 client for all.

Sounds good, @rravindran123 and I shall work on the singleton pattern implementation as a next step. Thanks

PR #419, opened 2023-10-19
refactor to create a single gitea client instance

vjayaramrh:


controllers/pkg/giteaclient/giteaclient.go
```diff
@@ -36,10 +37,25 @@ type GiteaClient interface {
...
 }
 
-func New(client resource.APIPatchingApplicator) GiteaClient {
-	return &gc{
-		client: client,
+var lock = &sync.Mutex{}
+
+var singleInstance *gc
+
+func New(client resource.APIPatchingApplicator, ctx context.Context) GiteaClient {
```
johnbelamaric, line 44: I guess my only question is whether this should be called "New" or something else. I wonder if there is a convention in Go for singletons. Otherwise LGTM (but need some unit tests).

vjayaramrh: > I guess my only question is whether this should be called "New" or something else. I wonder if there is a convention in Go for singletons. Otherwise LGTM (but need some unit tests).

`getInstance` would be better. Unit tests may require a mocking framework per my initial analysis, but am open for suggestions

henderiw, line 44: go best practice is to always add ctx as a first arg in a fn.

controllers/pkg/giteaclient/giteaclient.go
```diff
@@ -36,10 +37,25 @@ type GiteaClient interface {
 	Get() *gitea.Client
 }
 
-func New(client resource.APIPatchingApplicator) GiteaClient {
-	return &gc{
-		client: client,
+var lock = &sync.Mutex{}
```
henderiw, line 40: the mutex lock should be in the gate struct

controllers/pkg/giteaclient/giteaclient.go
```diff
@@ -36,10 +37,25 @@ type GiteaClient interface {
...
+
+var singleInstance *gc
+
+func New(client resource.APIPatchingApplicator, ctx context.Context) GiteaClient {
+	if singleInstance == nil {
```
henderiw, line 45: is it not better to have a flag like started and if set the started is not called

henderiw:
the get fn will also need a lock.

henderiw (changes requested):

controllers/pkg/giteaclient/giteaclient.go
vjayaramrh: Thanks for pointing out, will update the PR

controllers/pkg/giteaclient/giteaclient.go
vjayaramrh: Very similar to the original code, `Start(ctx)` is called only once when the instance is created for first time as can be seen in lines from L48 to L51 unless I misunderstood your comment.

vjayaramrh:
> the get fn will also need a lock.

Per my understanding of the code, `gc.giteaClient ` is set in the `Start` function which is invoked only once. The get fn that you are referring to returns the gc.giteaClient.  Since, it is not getting updated too often, my understanding is that there is not a need for lock in the get fn. Please correct if my understanding is incorrect. Thanks

controllers/pkg/giteaclient/giteaclient.go
henderiw: you can do either way, so keep as is.

controllers/pkg/giteaclient/giteaclient.go
henderiw: it is a flavour of how I do it, but your approach is ok

henderiw:
> > the get fn will also need a lock.
> 
> Per my understanding of the code, `gc.giteaClient ` is set in the `Start` function which is invoked only once. The get fn that you are referring to returns the gc.giteaClient. Since, it is not getting updated too often, my understanding is that there is not a need for lock in the get fn. Please correct if my understanding is incorrect. Thanks

I was too quick, you have a lock on the singleton which is also a valid approach. hence the get does not need a lock since it is done on the singleton

controllers/pkg/giteaclient/giteaclient.go
vjayaramrh: @johnbelamaric FYI I have added some basic boundary tests as a start

henderiw:
There is 1 more optimisation we can do in the logger. we can remove l log.Logr in the gate struct.

and use log := log.FromContext(ctx) and use log.Info iso r.l.Info.

vjayaramrh:
> There is 1 more optimisation we can do in the logger. we can remove l log.Logr in the gate struct.
> 
> and use log := log.FromContext(ctx) and use log.Info iso r.l.Info.

@henderiw Great suggestion, but I believe, it is better to do across all the impacted files as part of a separate PR
```
[openshift@fedora37 nephio]$ grep -nr logr.Logger
controllers/pkg/cluster/capi/capi.go:43:        l      logr.Logger
controllers/pkg/giteaclient/giteaclient.go:73:  l           logr.Logger
controllers/pkg/reconcilers/approval/reconciler.go:88:  l logr.Logger
controllers/pkg/reconcilers/bootstrap-packages/reconciler.go:93:        l logr.Logger
controllers/pkg/reconcilers/bootstrap-secret/reconciler.go:70:  l logr.Logger
controllers/pkg/reconcilers/generic-specializer/reconciler.go:93:       l logr.Logger
controllers/pkg/reconcilers/ipam-specializer/reconciler.go:91:  l logr.Logger
controllers/pkg/reconcilers/network/reconciler.go:131:  l       logr.Logger
controllers/pkg/reconcilers/network/watch_endpoint.go:36:       l      logr.Logger
controllers/pkg/reconcilers/network/watch_node.go:36:   l      logr.Logger
controllers/pkg/reconcilers/repository/reconciler.go:90:        l logr.Logger
controllers/pkg/reconcilers/token/reconciler.go:92:     l logr.Logger
controllers/pkg/reconcilers/vlan-specializer/reconciler.go:89:  l logr.Logger
controllers/pkg/specializer-reconciler/reconciler.go:68:        l logr.Logger
```
How about I open an issue and partner with a new comer looking to make beginner contributions to Nephio project and deliver it as part of a separate PR?

controllers/pkg/giteaclient/giteaclient.go
```diff
@@ -36,10 +37,33 @@ type GiteaClient interface {
...
 }
 
-func New(client resource.APIPatchingApplicator) GiteaClient {
-	return &gc{
-		client: client,
+var lock = &sync.Mutex{}
+
+var singleInstance *gc
+
+func GetGiteaClientInstance(ctx context.Context, client resource.APIPatchingApplicator) (GiteaClient, error) {
```
johnbelamaric, line 44: I think just "GetClient" is more idiomatic - we already know it's a Gitea client from the package, and "Get" hints that it's retrieving rather than creating, so "Instance" seems redundant.

vjayaramrh: Sounds good

johnbelamaric (approved):
/approve
/lgtm

PR #421, opened 2023-10-20
WIP - Add support for OpenShift cluster provider

adetalhouet:
The following PR https://github.com/nephio-project/nephio-example-packages/pull/86 adds the proper RBAC to the nephio bootstrap controller


johnbelamaric:
My only question here is whether this should instead be part of a separate, OpenShift set of controllers.

If you look, for example, at K8s, they started with kube controller manager that embedded cloud-specific things, just like this. But eventually they realized maintaining that was hard - you end up with a binary with many, many, many dependencies, and those dependencies are updated on different schedules for different providers. Instead, they spent literally *years* splitting all that code out into separate, per cloud-provider controller managers.

My guess is we will find ourselves in a similar situation, where there are various controllers we need for each environment; those can all be bundled up into a vendor-specific controller manager where the user can run that controller-manager if they want to work with that environment. I would rather we start with the right set up for our code base, rather than a major refactor later on.

So, I would suggest:
- If there are common bits/interfaces, they will live in the Nephio repo in a "pkg" directory.
- The current bootstrap controller becomes the "capi" bootstrap controller. It will continue to live in the nephio-project/nephio repository.
- We create a separate, openshift controller manager to house all of the openshift-specific code. It can use the shared interfaces and routines from the pkg directory so we don't duplicate a bunch of code.
- The openshift controller manager will be a separate binary. I think it can still live in the "nephio-project/nephio" repo for now, but ideally it would split out into a separate openshift-specific repository (within nephio-project, ideally, so it can be part of our distribution).

In the openshift environment (either for the mgmt cluster or even just for workload clusters), users would enable the openshift-controller-manager.

WDYT?

I have a couple GCP controllers (well, one that I use, but there are others) that are part of Porch right now. Ultimately I want to do the same with those.

controllers/pkg/cluster/cluster.go
```diff
@@ -21,20 +21,30 @@ import (
...
+		case "cluster.x-k8s.io/secret":
 			return &capi.Capi{Client: r.Client, Secret: secret}, true
+		case "Opaque":
+			// In case of OpenShift, the secret will have the following name
+			// $clusterName-admin-kubeconfig and the label hive.openshift.io/secret-type: kubeconfig
+			val, ok := secret.Labels["hive.openshift.io/secret-type"]
+			if ok && val == kubeconfig {
+				return &openshift.OpenShift{Client: r.Client, Secret: secret}, true
+			}
 		}
```
liamfallon, line 48: Is there another approach here rather than hard coding the Openshift specific call? Perhaps a way of configuring this so that the label and the function to call is read from config/metadata would be a better approach.

efiacor:
Can we get a status update on this PR @adetalhouet ?

efiacor:
> > Can we get a status update on this PR @adetalhouet ?
> 
> What kind of update would you like? I don't have the time to test this between now and mid-Feb. So if this is a gating factor, then discard this PR for R2. I tested this back then when I submitted the PR. I'll have some time after mid-Feb.

Hi @adetalhouet , I was just trying to get some movement on the outstanding PRs really. Support for alternative cluster provisioning is definitely needed so maybe we mark as WIP and revisit in R-3

adetalhouet:
> > > Can we get a status update on this PR @adetalhouet ?
> > 
> > 
> > What kind of update would you like? I don't have the time to test this between now and mid-Feb. So if this is a gating factor, then discard this PR for R2. I tested this back then when I submitted the PR. I'll have some time after mid-Feb.
> 
> Hi @adetalhouet , I was just trying to get some movement on the outstanding PRs really. Support for alternative cluster provisioning is definitely needed so maybe we mark as WIP and revisit in R-3

OpenShift cluster provisioning works without this. This PR is simply to use the bootstrap sequence defined in Nephio, with the ultimate goal to install configsync and being able to then sync anything from there.
OpenShift portfolio has other means to do so that will be leveraged if this can't be merged. 

Given the tight timeline for R2 and the actual bandwidth I have, I doubt I can revisit this. I'll mark as WIP and put in draft for now.

efiacor:
> > > > Can we get a status update on this PR @adetalhouet ?
> > > 
> > > 
> > > What kind of update would you like? I don't have the time to test this between now and mid-Feb. So if this is a gating factor, then discard this PR for R2. I tested this back then when I submitted the PR. I'll have some time after mid-Feb.
> > 
> > 
> > Hi @adetalhouet , I was just trying to get some movement on the outstanding PRs really. Support for alternative cluster provisioning is definitely needed so maybe we mark as WIP and revisit in R-3
> 
> OpenShift cluster provisioning works without this. This PR is simply to use the bootstrap sequence defined in Nephio, with the ultimate goal to install configsync and being able to then sync anything from there. OpenShift portfolio has other means to do so that will be leveraged if this can't be merged.
> 
> Given the tight timeline for R2 and the actual bandwidth I have, I doubt I can revisit this. I'll mark as WIP and put in draft for now.

Thanks you sir :)

tliron:
@adetalhouet What do you think about getting this ready for R3?

adetalhouet:
After all, this is not really needed. Once Nephio will tackle cluster management / LCM, I'm sure this will be revisited. It is already possible to create and configure an OpenShift cluster using Nephio. This patch is the push the integration further, but I don't think it is needed at this point. Will re-open / investigate when we can there in Nephio.

PR #423, opened 2023-10-24
Enforcing the check of env value

rravindran123:
This PR checks the value of the Environmental value in addition to just checking its definition as it is currently implemented.


operators/nephio-controller-manager/main.go
```diff
@@ -204,14 +204,17 @@ func parseReconcilers(reconcilers string) []string {
...
 	if slices.Contains(reconcilers, "*") {
 		return true
 	}
 	if slices.Contains(reconcilers, reconciler) {
 		return true
 	}
-	if _, found := os.LookupEnv(fmt.Sprintf("ENABLE_%s", strings.ToUpper(reconciler))); found {
-		return true
+	if val, found := os.LookupEnv(fmt.Sprintf("ENABLE_%s", strings.ToUpper(reconciler))); found {
+		if val == "true" {
```
subh007, line 215: we could avoid extra check :
```
return val
```
:)

johnbelamaric:
/approve

yes, please

vjayaramrh (approved):
/lgtm

PR #426, opened 2023-10-31
Unit test coverage for DeleteRepo and upsertRepo functions in repository reconciler

liamfallon:


liamfallon:
/assign @vjayaramrh @rravindran123 @efiacor @lapentad 

Unit test coverage for a small function as a starting point.

controllers/pkg/giteaclient/giteaclient.go
```diff
@@ -137,3 +141,11 @@ func getClientAuth(secret *corev1.Secret) gitea.ClientOption {
 func (r *gc) Get() *gitea.Client {
 	return r.giteaClient
 }
+
+func (r *gc) GetMyUserInfo() (*gitea.User, *gitea.Response, error) {
```
vjayaramrh, line 145: What would be the goal for adding this method? If one invokes the Get() method, they all already have access to the client and could invoke all the functions the client object offers, right? Would be adding other functions that this client supports as methods as well in the future? THanks

controllers/pkg/giteaclient/giteaclient.go
```diff
@@ -137,3 +141,11 @@ func getClientAuth(secret *corev1.Secret) gitea.ClientOption {
...
+	return r.giteaClient.GetMyUserInfo()
+}
+
+func (r *gc) DeleteRepo(owner string, repo string) (*gitea.Response, error) {
```
vjayaramrh, line 149: Same as above

controllers/pkg/giteaclient/giteaclient.go
liamfallon: I have to introduce this function and the DeleteRepo function in order to make mocking possible in the reconciler. I could not write a mock in the reconciler because the reconciler code used the Get() function to get a reference to the implementation of the interface and then called the implementaiton directly, therefore short circuiting the interface.

What is done here in the PR is actually only the first step of what we need to do to refactor the reconciler so that we can use repositories other than gitea. The reconciler should only call functions on the interface and then there can be different implementations of that interface for Gitea and Git.

In fact, perhaps we should remove the Get() method from the interface altogether to prevent direct access to the implementation in users of the interface.

johnbelamaric: So, it sounds like this refactoring is aiming to have a single reconciler for various Git providers. This is the same  pattern and concerns as what we discussed last week (see https://github.com/nephio-project/nephio/pull/421#issuecomment-1775707111).

liamfallon: Actually having different reconcilers for different repo providers would be a good approach. In a particular installation you can configure the system to use whichever one you want to run.

liamfallon: After yesterday's SIG-Automation meeting, we will discuss the restructuring/refactoring separately. Can we review this PR as a pure Unit Test review? The refactoring is done just to aid unit testing.

vjayaramrh (approved):
/lgtm
@liamfallon overall, the approach looks good to me,
We may also want to do a POC using https://pkg.go.dev/github.com/stretchr/testify/mock to generate mocks and see if it could be better. I have used it and found it useful to create mocks and write unit tests. We may have to come up with training material to make it easy for community members to adopt if we think it is good.
Also, does it make sense to rename the directory from repository to gitea-repository as the code is specific to gitea?

liamfallon:
> /lgtm @liamfallon overall, the approach looks good to me, We may also want to do a POC using https://pkg.go.dev/github.com/stretchr/testify/mock to generate mocks and see if it could be better. I have used it and found it useful to create mocks and write unit tests. We may have to come up with training material to make it easy for community members to adopt this if we think it is good. Also, does it make sense to rename the directory from repository to gitea-repository as the code is specific to gitea?

Yes, we should also try to do a POC to try the mock generation.

It probably would be a good idea to rename this to gitea-repository but maybe let's discuss this in the community and take it in another PR

liamfallon:
Can I have some reviews on this PR please?

controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -119,7 +119,7 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		// Delete the repo from the git server
 		// when successful remove the finalizer
 		if cr.Spec.Lifecycle.DeletionPolicy == commonv1alpha1.DeletionDelete {
-			if err := r.deleteRepo(ctx, giteaClient, cr); err != nil {
+			if err := r.deleteRepo(ctx, r.giteaClient, cr); err != nil {
```
johnbelamaric, line 122: Not part of this PR (no need to address), but line 108 is weird, why doesn't it just check r.giteaClient directly? I don't think the Get() method should be needed. But maybe a follow up PR (with the change in this line that will become a short PR).

johnbelamaric (approved):
/approve

controllers/pkg/reconcilers/repository/reconciler.go
liamfallon: Yes, I agree. I'll do this in the next PR.

PR #431, opened 2023-11-02
Remove config-sync dependency from bootstrap controller

efiacor:
Remove config-sync lookup from bootstrap controller.
Minor refactoring.
Add unit test coverage.


.gitignore
```diff
@@ -20,8 +20,9 @@ vendor
 # Tox environments
 .tox/
 *.dic
-/.vscode
+**/.vscode/
```
henderiw, line 23: do we need to update in this PR?

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
 type reconciler struct {
 	client.Client
 	porchClient client.Client
-
-	l logr.Logger
+	log         logr.Logger
```
henderiw, line 85: we can remove this and use the log.FromContext.

so do this:
log := log.FromContext(ctx)
iso
r.log =  log.FromContext(ctx)

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
 }
 
 func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	r.l = log.FromContext(ctx)
-	cr := &porchv1alpha1.PackageRevision{}
-	if err := r.Get(ctx, req.NamespacedName, cr); err != nil {
+	r.log = log.FromContext(ctx)
+	packageRevision := &porchv1alpha1.PackageRevision{}
+	if err := r.Client.Get(ctx, req.NamespacedName, packageRevision); err != nil {
```
henderiw, line 91: We are discussing this in another thread but this means we have to change the code for every controller rather than just copying things. When I started I did this, but if you write many controllers you get bored by it. This is DDD.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
+			r.log.Error(err, msg)
 			return ctrl.Result{}, errors.Wrap(resource.IgnoreNotFound(err), msg)
 		}
 		return ctrl.Result{}, nil
 	}
 
 	// check if the packagerevision is part of a staging repository
 	// if not we can ignore this package revision
-	stagingPR, err := r.IsStagingPackageRevision(ctx, cr)
+	stagingPR, err := r.IsStagingPackageRevision(ctx, r.porchClient, packageRevision.Spec.RepositoryName)
```
henderiw, line 104: it is not good practice to add a variable that is already in your context also as a parameter.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
 	if err != nil {
 		msg := "cannot list repositories"
-		r.l.Error(err, msg)
+		r.log.Error(err, msg)
```
henderiw, line 107: can be optimised as I mentioned before

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
 		return ctrl.Result{}, errors.Wrap(err, msg)
 	}
-	if stagingPR && porchv1alpha1.LifecycleIsPublished(cr.Spec.Lifecycle) {
-		r.l.Info("reconcile package revision")
-		resources, namespacePresent, err := r.getResources(ctx, req)
+	if stagingPR && porchv1alpha1.LifecycleIsPublished(packageRevision.Spec.Lifecycle) {
+		r.log.Info("reconcile package revision")
```
henderiw, line 111: can be optimised as I mentioned before

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
-		// first resource in the resource list
 		if len(resources) > 0 {
-			clusterName, ok := resources[0].GetAnnotations()[clusterNameKey]
+			// we expect the clusterName to be applied to all resources in the
+			// package revision resources, so we find the cluster name by looking at the
+			// first resource in the resource list
+			clusterName, ok := resources[0].GetAnnotations()["nephio.org/cluster-name"]
 			if !ok {
-				r.l.Info("clusterName not found",
+				r.log.Info("clusterName not found",
```
henderiw, line 125: can be optimised as I mentioned before

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
 					"resource", fmt.Sprintf("%s.%s.%s", resources[0].GetAPIVersion(), resources[0].GetKind(), resources[0].GetName()),
 					"annotations", resources[0].GetAnnotations())
 				return ctrl.Result{}, nil
 			}
-			// we need to find the cluster client
-			secrets := &corev1.SecretList{}
-			if err := r.List(ctx, secrets); err != nil {
-				msg := "cannot list secrets"
-				r.l.Error(err, msg)
+			clusterSecret, err := r.GetClusterSecret(ctx, r.Client, clusterName)
```
henderiw, line 130: you should not have to put r.Client as a param

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
+					return ctrl.Result{RequeueAfter: 30 * time.Second}, errors.Wrap(err, msg)
+				}
+				if !ready {
+					r.log.Info("cluster not ready")
+					return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+				}
+				// install the resources to the cluster
+				for _, resource := range resources {
+					resource := resource // required to prevent gosec warning: G601 (CWE-118): Implicit memory aliasing in for loop
+					r.log.Info("install manifest", "resource", fmt.Sprintf("%s.%s.%s", resource.GetAPIVersion(), resource.GetKind(), resource.GetName()))
```
henderiw, line 151: can be optimised as I mentioned before

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
 			}
 		}
 	}
 	return ctrl.Result{}, nil
 }
 
-func (r *reconciler) IsStagingPackageRevision(ctx context.Context, cr *porchv1alpha1.PackageRevision) (bool, error) {
-	repos := &porchconfigv1alpha1.RepositoryList{}
-	if err := r.porchClient.List(ctx, repos); err != nil {
+func (r *reconciler) GetClusterSecret(ctx context.Context, client client.Client, clusterName string) (corev1.Secret, error) {
```
henderiw, line 168: don't put the client as a param

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
```diff
@@ -89,144 +82,135 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c a
...
+	for _, secret := range secrets.Items {
+		if strings.Contains(secret.GetName(), clusterName) {
+			clusterSecret = secret
+			break
+		}
+	}
+	return clusterSecret, nil
+}
 
+func (r *reconciler) IsStagingPackageRevision(ctx context.Context, porchClient client.Client, repositoryName string) (bool, error) {
```
henderiw, line 184: don't put porchClient client.Client here

default-go-test.mk
```diff
@@ -24,11 +25,13 @@ unit: test
 test: ## Run unit tests (go test)
 ifeq ($(CONTAINER_RUNNABLE), 0)
 		$(CONTAINER_RUNTIME) run -it -v "$(CURDIR):/go/src" -w /go/src docker.io/library/golang:${GO_VERSION}-alpine3.17 \
-         sh -e -c "go test ./... -v -coverprofile ${TEST_COVERAGE_FILE}; \
+         sh -e -c "go test ./... -v -coverprofile ${TMP_TEST_COVERAGE_FILE}; \
```
henderiw, line 28: do we need to add this file?

controllers/pkg/reconcilers/bootstrap-packages/fake_client.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2023 The Nephio Authors
```
henderiw, line 1: this file should move the porch client library iso in the bootstrap pkg

henderiw (changes requested):

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: r.log is being used in the getPrResources and sub function so how to access it from there? Pass it as a param?

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: I take your point but for me readability is more important. As a human, who is much more prone to errors, code should be easy to follow. In this context, the var name may be obvious to most, but I don't see the gain in not naming it descriptively.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: I will admit, I struggled to unit test these functions. Passing it as a parameter was the obvious way to allow me to mock the client. If there is an alternative way, without passing it in, please do advise when you have time.
I looked at trying to use the fake.go but because I needed to mock actual return values, I didn't get very far with it.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Could you double check how to access the logger from the getPrResources and it's sub function?

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Same issue with passing the mock client.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: As above

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Same issue with passing the mock client.

default-go-test.mk
efiacor: I did this to strip the fake_client.go from the coverage report. I guess I could name it with a _test.go suffix.
The report also doesn't really give a proper reflection of the overall coverage. Do we need an _test.go in each reconciler dir so that a report is produced?

controllers/pkg/reconcilers/bootstrap-packages/fake_client.go
efiacor: I did consider this but thought this fake is quite specific to this controller. If another controller requires a mock client, their impl would differ so should this not live next to the controller it's mocking the client for?

.gitignore
efiacor: May be a noob and or IDE user mistake but I am coding from the pkg dir in vscode. This produces the local .vscode file which gets picked up by git. I can remove it manually before committing but just made sense to add it now.

henderiw: if we need to change it we should do it in another PR, that asks for this change

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: if you pass the context you can do the same there. The controller SDK has this as a mechanism. You pass the logger via the ctx and you consume it through the ctx.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: ok we differ in opinion here

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: you can create a new reconciler in which you pass this info. So build a reconciler in the Mock that has the Mock client and you call the fn like that.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: via the ctx

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: see above

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: see above

default-go-test.mk
henderiw: if we have to change the report we should do this in another PR. in go you normally put a _test.go file next to the functions. Now if we can improve the reporting we can do but do in another PR.

controllers/pkg/reconcilers/bootstrap-packages/fake_client.go
henderiw: I believe it is better to make it generic, so other people can consume it. I did the same for ipam and clan an tried to build in generic capabilities to satisfy multiple consumers. We otherwise have to maintain all this code and it is all over the place.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
henderiw: so here is my suggestion, if you want to debate this put this in another PR and we can debate it there. In general it is better to do a PR that changes 1 thing iso multiple things, since you see that you might get stuck on certain aspects.

.gitignore
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

default-go-test.mk
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/fake_client.go
efiacor: I will need some assistance here. The fake is for any controller runtime client really so not porch specific. I have hacked a solution to load the response data from files but needs to be refined/refactored.

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

controllers/pkg/reconcilers/bootstrap-packages/reconciler.go
efiacor: Done

henderiw:
My suggestion on this PR is the following:
- split this PR in 2 parts
1. the changes to remove config sync dependency -> they are good to go from my pov
2. the changes to add tests -> these should go to the porch client as we are not testing the reconciler but we are testing the porch client.

efiacor:
> My suggestion on this PR is the following:
> 
> * split this PR in 2 parts
> 
> 1. the changes to remove config sync dependency -> they are good to go from my pov
> 2. the changes to add tests -> these should go to the porch client as we are not testing the reconciler but we are testing the porch client.

Hi @henderiw ,
I strongly disagree with this plan. I added a test to cover the logic contained in the [IsStaging function](https://github.com/nephio-project/nephio/blob/main/controllers/pkg/reconcilers/bootstrap-packages/reconciler.go#L198).
Stubbing the response from the server is the only way to do this. If we simply "test" the porch client then the calling logic doesn't get tested.
The lack of testing on the code is really bad practice. If the feeling of the project is that unit test is useless/not required then I will go with that but this has to be agreed at project level.

henderiw:
I am not pushing against test I am just saying the tests you are doing are about testing the porch client and they should move there.

henderiw:
what I am proposing is putting them here:

https://github.com/nephio-project/nephio/tree/main/controllers/pkg/porch/client

henderiw:
@efiacor could we move this test to this place?
https://github.com/nephio-project/nephio/tree/main/controllers/pkg/porch/client

than we can get this closed. thx

efiacor:
> @efiacor could we move this test to this place? https://github.com/nephio-project/nephio/tree/main/controllers/pkg/porch/client
> 
> than we can get this closed. thx

Hi @henderiw ,
I removed the tests from this PR and plan to re add some form of testing in a later PR

henderiw:
ok we can also do in 2 steps. Thx @efiacor

PR #433, opened 2023-11-03
Add unit tests for porch.condition

efiacor:
Add unit tests to condition module.
94%

### DIFF
--- a/controllers/pkg/porch/condition/condition.go
+++ b/controllers/pkg/porch/condition/condition.go
@@ -69,14 +69,3 @@ func PackageRevisionIsReady(readinessGates []porchv1alpha1.ReadinessGate, condit
 
 	return true
 }
-
-// HasReadinessGate checks if a specific gate exists
-func HasReadinessGate(gates []porchv1alpha1.ReadinessGate, gate string) bool {
-	for i := range gates {
-		g := gates[i]
-		if g.ConditionType == gate {
-			return true
-		}
-	}
-	return false
-}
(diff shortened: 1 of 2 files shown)


controllers/pkg/porch/condition/condition.go
```diff
@@ -69,14 +69,3 @@ func PackageRevisionIsReady(readinessGates []porchv1alpha1.ReadinessGate, condit
 
 	return true
 }
-
-// HasReadinessGate checks if a specific gate exists
-func HasReadinessGate(gates []porchv1alpha1.ReadinessGate, gate string) bool {
```
johnbelamaric, line 74: You removed this because it's not used anywhere?

controllers/pkg/porch/condition/condition_test.go
```diff
@@ -106,3 +106,49 @@ func TestHasSpecificTypeConditions(t *testing.T) {
...
 	}
 }
+
+func TestPackageRevisionIsReady(t *testing.T) {
+	cases := map[string]struct {
+		conds    []porchv1alpha1.Condition
+		readyGates []porchv1alpha1.ReadinessGate
+		want bool
+	}{
+		"Ready": {
```
johnbelamaric, line 116: Can you add some false conditions in here to show that it's only the one listed in the readiness gates that matters?

controllers/pkg/porch/condition/condition_test.go
```diff
@@ -106,3 +106,49 @@ func TestHasSpecificTypeConditions(t *testing.T) {
 		})
 	}
 }
+
+func TestPackageRevisionIsReady(t *testing.T) {
+	cases := map[string]struct {
+		conds    []porchv1alpha1.Condition
```
vjayaramrh, line 112: how about tests for conds and readyGates being an empty list?

controllers/pkg/porch/condition/condition.go
johnbelamaric: :man_shrugging: if tests pass let's try to get rid of it and see...

PR #438, opened 2023-11-13
Optimize logging for repo controller

efiacor:
Log from context directly.

### DIFF
--- a/controllers/pkg/reconcilers/repository/reconciler.go
+++ b/controllers/pkg/reconcilers/repository/reconciler.go
@@ -41,7 +41,7 @@ func init() {
 }
 
 const (
-	finalizer = "infra.nephio.org/finalizer"
+	finalizer       = "infra.nephio.org/finalizer"
 	errUpdateStatus = "cannot update status"
 )
 
--- a/controllers/pkg/reconcilers/repository/reconciler.go
+++ b/controllers/pkg/reconcilers/repository/reconciler.go
@@ -22,7 +22,6 @@ import (
 	"reflect"
 
 	"code.gitea.io/sdk/gitea"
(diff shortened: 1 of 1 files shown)


controllers/pkg/reconcilers/repository/reconciler.go
```diff
@@ -42,9 +41,6 @@ func init() {
 }
 
 const (
-	finalizer = "infra.nephio.org/finalizer"
```
henderiw, line 45: we could keep these here since it is good practice to initialise the const. So the motivation is to be apply to copy paste the code below and the dev just has to update the const. The rest will flow.

henderiw (changes requested):

controllers/pkg/reconcilers/repository/reconciler.go
efiacor: The err isn't used at all so will remove.
The finalizr const is used once so doesn't really require a const but will revert anyway.

efiacor: > we could keep these here since it is good practice to initialise the const. So the motivation is to be apply to copy paste the code below and the dev just has to update the const. The rest will flow.

The err isn't used at all so will remove.
The finalizr const is used once so doesn't really require a const but will revert anyway.

henderiw: yes err can eb removed, but in another controller you just need to change the const at the top

efiacor: > yes err can eb removed, but in another controller you just need to change the const at the top

Understood. Done

PR #446, opened 2023-11-27
Add mockery test case to token reconciler

efiacor:
Add mock nephio gitea client.
Add example test case fro deketeToken function.
Anti pattern arises in relation to the gitea client wrapper as it's also being used by the repo reconciler.


efiacor:
@vjayaramrh 
Using your example as a base I added a simple test to the token reconciler deleteToker function.
Some bits would need some clarification regarding the method signature and how the responses can be manipulated but over all, pretty straigh forward.

controllers/pkg/reconcilers/token/reconciler_test.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := &reconciler{
+				APIPatchingApplicator: tt.fields.APIPatchingApplicator,
+				giteaClient:           tt.fields.giteaClient,
+				finalizer:             tt.fields.finalizer,
+			}
+			// The below block being setup and processing of mocks before invoking the function to be tested
```
vjayaramrh, line 79: Looks like the block of code from L85 to L91 can be captured in a function to be reused by every test case that uses mockery, right? What do you think?

liamfallon: Maybe we can leave it here for now and maybe refactor later. I'd like to get this PR merged so that I can refactor the Repository unit tests to use Mockery.

PR #453, opened 2023-11-30
Porch to nephio

johnbelamaric:
As discussed in #450, this PR imports the Porch server-side code and makes it build using our standard `make docker-build` target.

This includes code up through https://github.com/kptdev/kpt/pull/4097. There will be a new release of Porch today that should be equivalent to this code. I think we would be best off if we build the Porch binaries with this code and shipped these images in the Nephio Docker Hub for R2, but I will leave that up to @gvbalaji and SIG Release.

@radoslawc once this merges, it would be great to get automated builds of the docker images in here.

Note that post-merge, we will remove a lot of code.


johnbelamaric:
@radoslawc @gvbalaji one error in the lichen scan:
```2023/11/30 22:04:41 1 error occurred:
	* github.com/opencontainers/go-digest@v1.0.0: not allowed - non-permitted licenses: [CC-BY-SA-4.0]
```

Can we check if that license is OK?

johnbelamaric:
In particular, after deleting all the code we don't want to keep...

porch/.vscode/launch.json
efiacor, line 1: Do we want to keep this or?

porch/api/doc.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2022 The kpt Authors
```
efiacor, line 1: Would we need to update the headers?

porch/cmd/porch/main.go
```diff
@@ -0,0 +1,135 @@
...
+	"k8s.io/component-base/cli"
+	"k8s.io/klog/v2"
+)
+
+func main() {
+	code := run()
+	os.Exit(code)
+}
+
+func run() int {
```
efiacor, line 45: init?

efiacor (commented):
A lot here so only skimmed it. The docs for me would be the top priority so that others can get up to speed quickly.

johnbelamaric:
Closing in favor of https://github.com/nephio-project/porch/pull/2

PR #456, opened 2023-12-06
Only overwrite the specific annotations

henderiw:
We were resetting all annotations instead we update the specific once that became specific to the remote cluster


liamfallon:
It would be great to get the formatting changes and the code changes in separate PRs. It would make it easier to spot the changes.

efiacor (approved):

PR #441, opened 2023-11-15
sample code that uses mockery and testify for writing unit tests

vjayaramrh:
The goal of this PR is to demonstrate usage of testify (https://pkg.go.dev/github.com/stretchr/testify) and mockery (https://github.com/vektra/mockery/tree/master) for writing unit tests.
Will be using comments in the PR to help folks come upto speed on using testify and mockery.

Steps to try out:
=============
- a)Pull the PR to your dev environment, b)delete the mock_GiteaClient.go file c)execute `make install-mockery` d)execute `make generate-mocks` to verify `mock_GiteaClient.go` is created.

References:
===========
Mockery -> https://vektra.github.io/mockery/latest/#why-mockery
Testify ->  https://pkg.go.dev/github.com/stretchr/testify
Installing Testify --> https://pkg.go.dev/github.com/stretchr/testify#readme-installation

PR #461, opened 2023-12-14
Converted Repository operator tests to Mockery

liamfallon:
This PR uses the approach and code described by @vjayaramrh in https://github.com/nephio-project/nephio/pull/441 to replace the unit tests for the Repository operator. This PR replaces all the unit tests with Mockery implementations.


controllers/pkg/reconcilers/repository/reconciler_test.go
```diff
@@ -16,187 +16,260 @@ package repository
...
-func (gc NephioGiteaClientMock) Get() *gitea.Client {
-	return nil
-}
-
-func (gc NephioGiteaClientMock) Start(ctx context.Context) {
-}
-
-func (gc NephioGiteaClientMock) GetMyUserInfo() (*gitea.User, *gitea.Response, error) {
-	return &gitea.User{}, &gitea.Response{}, gc.myUserInfoError
+type mockHelper struct {
```
vjayaramrh, line 33: Suggestion: it may be a good idea to create a separate directory called testing and defined the mockHelper struct there and then could be imported by other packages that will use mockery for unit testing

controllers/pkg/reconcilers/repository/reconciler_test.go
```diff
@@ -16,187 +16,260 @@ package repository
...
+			initMockeryMocks(&tt)
+
+			if err := r.deleteRepo(tt.args.ctx, tt.args.giteaClient, tt.args.cr); (err != nil) != tt.wantErr {
+				t.Errorf("deleteRepo() error = %v, wantErr %v", err, tt.wantErr)
+			}
 		})
 	}
 }
+
+func initMockeryMocks(tt *repoTest) {
```
vjayaramrh, line 262: Suggestion: This function could be moved to a the testing folder as well to be reused by others.

liamfallon:
I need to split the mockeryutils module out into a separate PR so that it can go in first and avoid the missing ref from the reconciler.

liamfallon:
We need to wait for this PR to be merged
https://github.com/nephio-project/nephio/pull/468

liamfallon:
@vjayaramrh @efiacor This is ready for review now.

.mockery.yaml
```diff
@@ -0,0 +1,6 @@
+packages:
+  github.com/nephio-project/nephio/controllers/pkg/giteaclient:
+    interfaces:
+      GiteaClient:
+        config:
+          dir: "{{.InterfaceDir}}"
```
efiacor, line 6: Not sure on the best way to fix this one but..
The mock file gets picked up by the coverage report. ie 0%
Using the config in mockery we can name the file with a _test suffix, but for me this breaks the import in the test.go.
Alternatively, which is what I plan to do for say the controller-runtime client mock, is to output to a "mocks" dir under controller/pkg and maybe exclude that in the report. Or we suffix them with _test and add the logic for that in the coverage cmd.
Thoughts?

testing/mockeryutils/Makefile
efiacor, line 1: Is this needed?

testing/mockeryutils/Makefile
```diff
@@ -1,25 +1,20 @@
...
-#  distributed under the License is distributed on an "AS IS" BASIS,
-#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#  See the License for the specific language governing permissions and
-#  limitations under the License.
-
-.PHONY: all
-all: fmt test
+IMAGE_TAG ?= latest
+REGISTRY ?= docker.io/nephio
+IMAGE_NAME ?= ipam-fn
```
efiacor, line 3: Is this correct? Do we need a Makefile here?

.mockery.yaml
liamfallon: Can we take this discussion separately? Would it be OK to bring this PR in? I'll look at this in a subsequent PR.

testing/mockeryutils/Makefile
liamfallon: Yes, if it's not here the build fails. The main Makefile expects a Makefile in every submodule that contains "go" files.

testing/mockeryutils/Makefile
liamfallon: Yes, if it's not here the build fails. The main Makefile expects a Makefile in every submodule that contains "go" files.

liamfallon: Thanks @efiacor , I used this Makefile

controllers/pkg/reconcilers/repository/reconciler_test.go
liamfallon: Done.

controllers/pkg/reconcilers/repository/reconciler_test.go
liamfallon: Done.

liamfallon:
> I need to split the mockeryutils module out into a separate PR so that it can go in first and avoid the missing ref from the reconciler.

Done

liamfallon:
@s3wong @tliron @johnbelamaric @vjayaramrh Can we get reviews on this please?

Issue #373, opened 2023-09-11
Run integration tests  prior to merging PRs


gvbalaji:

gvbalaji:
In works. 
https://prow.nephio.io/?type=periodic

PR #486, opened 2024-01-15
Adding pre-submit images build

radoslawc:
Dependency for pre-submit testing PRs:  https://github.com/nephio-project/nephio/issues/373

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -77,3 +77,363 @@ presubmits:
       - image: nephio/gotests:1671656214984921088
         command:
         - "/usr/local/bin/lichen.sh"
+
+  - name: build-push-pre-image-ueransim-deploy-fn
+    cluster: default
+    always_run: true
+    branches:
+    - "main"
+    annotations:
+      description: Build and Push Presubmit Image to Hub
+    decorate: true
+    decoration_config:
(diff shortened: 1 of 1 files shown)


.prow.yaml
```diff
@@ -77,3 +77,363 @@ presubmits:
       - image: nephio/gotests:1671656214984921088
         command:
         - "/usr/local/bin/lichen.sh"
+
+  - name: build-push-pre-image-ueransim-deploy-fn
```
efiacor, line 81: Is there any way in prow to define a job template that could be reused by all these?

radoslawc: @efiacor there is templating implemented, but extent of it doesn't cover everything we need. If you see other projects prow configuration suffers from the same problem being enormous. For main configuration one can split it to single file per job for example, for inrepoconfig (.prow.yaml) that's not an option.

efiacor: Yes, the closest I found is this, which is huge - https://github.com/aws/eks-distro-prow-jobs/blob/main/docs/prowjobs.md
Maybe we could use plank to define some common stuff but don't know prow well enough to be honest.

gvbalaji (commented):
/approve

PR #427, opened 2023-10-31
Rename variable for code readability

subh007:
Renamed variable for code readability.
changed `cr` to `network`.

PR #487, opened 2024-01-16
Rename variable for code readability

subh007:
Renamed variable for code readability.

### DIFF
--- a/controllers/pkg/reconcilers/network/reconciler.go
+++ b/controllers/pkg/reconcilers/network/reconciler.go
@@ -136,8 +136,8 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	log := log.FromContext(ctx)
 	log.Info("reconcile", "req", req)
 
-	cr := &infrav1alpha1.Network{}
-	if err := r.Get(ctx, req.NamespacedName, cr); err != nil {
+	network := &infrav1alpha1.Network{}
+	if err := r.Get(ctx, req.NamespacedName, network); err != nil {
 		// if the resource no longer exists the reconcile loop is done
 		if resource.IgnoreNotFound(err) != nil {
 			log.Error(err, errGetCr)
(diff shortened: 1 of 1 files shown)


henderiw:
as commented before I we should not do this

subh007:
Hello @henderiw , 

Continuing the discussion from here : https://github.com/nephio-project/nephio/pull/427 
I checked with other reviewers and they agree that it will improve the code readability.
@tliron @liamfallon @efiacor

liamfallon:
I took a look at this style guide from Google for go:
https://google.github.io/styleguide/go/decisions#naming

Subhash's proposed change from `cr` to `network` complies with the guidelines and makes the code easier to read. In the Nephio code, we are dealing with multiple CRs in multiple modules and it's not easy to keep track of where we are in our mental map of the code. Any help we can get with this is welcome.

Now if the variable had been named  something like `currentNetworkReconcilerCustomRssource`, that would be a different matter but `network` is concise and readable. This change is also aligned with coding guidelines I have seen in many languages over the years.

**_Abstract from Style Guide_**
 In general:
- Single-word names like `count` or `options` are a good starting point.
- Additional words can be added to disambiguate similar names, for example `userCount` and `projectCount`.
- Do not simply drop letters to save typing. For example `Sandbox` is preferred over `Sbx`, particularly for exported names.
- Omit [types and type-like words](https://google.github.io/styleguide/go/decisions#repetitive-with-type) from most variable names.
 o For a number, `userCount` is a better name than `numUsers` or `usersInt`.
 o For a slice, `users` is a better name than `userSlice`.
 o It is acceptable to include a type-like qualifier if there are two versions of a value in scope, for example you might have an input stored in `ageString` and use `age` for the parsed value.
- Omit words that are clear from the [surrounding context](https://google.github.io/styleguide/go/decisions#repetitive-in-context). For example, in the implementation of a `UserCount` method, a local variable called `userCount` is probably redundant; `count`, `users`, or even `c` are just as readable.

efiacor:
> Hello @henderiw ,
> 
> Continuing the discussion from here : #427 I checked with other reviewers and they agree that it will improve the code readability. @tliron @liamfallon @efiacor

Hi @subh007 , we should keep the orig PR to preserve the review comments. Can we reopen and rebase it please?
@henderiw can we go with this approach instead of using the generic "cr". For most with k8s and go knowledge, cr equals Custom Resource. That's logical. But I don't see the benefit if it obscures the readability for new devs. Especially if the purpose is primarily to ease the reuse (copy & paste) of the code.

subh007:
> > Hello @henderiw ,
> > Continuing the discussion from here : #427 I checked with other reviewers and they agree that it will improve the code readability. @tliron @liamfallon @efiacor
> 
> Hi @subh007 , we should keep the orig PR to preserve the review comments. Can we reopen and rebase it please? 

reopened the #427

henderiw:
I believe you don't get my point. If you write multiple controllers you will see that the proposed change does not make sense. Moreover if you write a generic library you don't even know what the variable means, so I advocate for this approach and as a result you get more reusable code.

What sub bash proposed I did in the early days and changed the approach as a lot becomes boilerplate and the meaning is irrelevant.

liamfallon:
@subh007 @henderiw I'm going to close this PR as it's been inactive for a long time. Please feel free to reopen.

PR #490, opened 2024-01-18
Add vscode workspace

efiacor:


liamfallon:
@efiacor  I'm not sure if this should be in the in the checked in code, it's a vscode thing. Maybe add an entry in the .gitignore so that it gets ignored on checkins?

PR #494, opened 2024-01-25
De duplication logic to NF Deploy Fn Param Ref

aakashchan:
These changes will ensure we have nf deploy fn to apply de dupulication logic to NF Deploy Param Ref

1. Changes to Add Dependency to check if it already exist before adding. Thanks @gvbalaji for the code snippet in chat.
2. Added test cases to handle dependency, if same file is present multiple times. Its not in our use case, but its better to add that too.
3. Added changes to pipeline tests to ensure, if I run the NF Deploy Fn multiples after that, it doesn't break the idempotency principle.

Docker image I used for testing - [docker.io/lostbrain/nfdeployfn:test-5](http://docker.io/lostbrain/nfdeployfn:test-5)

Issue #493, opened 2024-01-23
Duplicate entries after rendering the final NFDeployment template


arora-sagar:
When trying the latest `nfdeploy-fn` there are duplicate entries in `parametersRefs` field. I think we should do a check in lists to avoid duplicate entries. 

```
apiVersion: workload.nephio.org/v1alpha1
kind: NFDeployment
metadata: # kpt-merge: smf-example/smf-example
  name: smf-regional
  namespace: free5gc-cp
  annotations:
    internal.kpt.dev/upstream-identifier: 'workload.nephio.org|NFDeployment|smf-example|smf-example'
spec:
  provider: smf.free5gc.io
  interfaces:
  - name: n4
    ipv4:
      address: 172.1.1.254/24
      gateway: 172.1.1.1
    vlanID: 5
  networkInstances:
  - name: vpc-internal
    interfaces:
    - n4
  parametersRefs:
  - name: smf-regional-upf-edge01
    apiVersion: ref.nephio.org/v1alpha1
    kind: Config
  - name: smf-regional-upf-edge01
    apiVersion: ref.nephio.org/v1alpha1
    kind: Config
  - name: smf-regional-upf-edge01
    apiVersion: ref.nephio.org/v1alpha1
    kind: Config
  - name: smf-regional-upf-edge02
    apiVersion: ref.nephio.org/v1alpha1
    kind: Config
  - name: smf-regional-upf-edge02
    apiVersion: ref.nephio.org/v1alpha1
    kind: Config
  - name: smf-regional-upf-edge02
    apiVersion: ref.nephio.org/v1alpha1
    kind: Config
  capacity:
    maxDownlinkThroughput: "0"
    maxNFConnections: 5
    maxSessions: 500
    maxUplinkThroughput: "0"
```

__To repoduce__

Deploy free5gc smf function using R2 test-infra main branch.

electrocucaracha:
[This change](https://github.com/nephio-project/catalog/pull/28) removes the repeated `nfdeploy-fn` call which was affecting the process to generate SMF configuration file, but it's not fixing the duplication of parametersRefs

arora-sagar:
@tliron @electrocucaracha Should we keep it open till the time it is fixed?

tliron:
I think this is still open. We reverted the fix, so `nfdeploy-fn` is still not idempotent.

PR #525, opened 2024-02-14
De duplication logic to NF Deploy Fn Param Ref (#494)

aakashchan:
Solves #493 
These changes will ensure we have nf deploy fn to apply de dupulication logic to NF Deploy Param Ref

1. Changes to Add Dependency to check if it already exist before adding. Thanks @gvbalaji for the code snippet in chat.
2. Added test cases to handle dependency, if same file is present multiple times. Its not in our use case, but its better to add that too.
3. Added changes to pipeline tests to ensure, if I run the NF Deploy Fn multiples after that, it doesn't break the idempotency principle.

### DIFF
--- a/krm-functions/nfdeploy-fn/fn/function.go
+++ b/krm-functions/nfdeploy-fn/fn/function.go
@@ -259,6 +259,11 @@ func (f *NfDeployFn) UpdateResourceFn(nfDeploymentObj *fn.KubeObject, objs fn.Ku
 		return nil, err
 	}
 
+	// add all the existing param ref to the struct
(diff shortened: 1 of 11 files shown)


efiacor:
Hi @aakashchan , is this change still relevant?

tliron:
@aakashchan bump :)

aakashchan:
@radoslawc I'm waiting for you confirm. In my testing environment, I didn't face issue and not able to replicate.

efiacor:
Looks like this change is needed. Please review and progress.

PR #511, opened 2024-02-12
Add test coverage to token rec

efiacor:
Refactor existing test to use mockery helper functions
Add controller-runtime mock

PR #526, opened 2024-02-14
Refactor support for Mockery in the makefiles

liamfallon:
The support for mockery is moved from all other makefiles and .mk files into the `default-mockery.mk` file. Mockery related code in other makefiles and .mk files is removed.

The included `default-go.mk` file now includes `default-mockery.mk`.

There are two targets in `default-mockery.mk`:

1. `install-mockery`: Installs mockery in docker or locally if docker is not available
2. `generate-mocks`: Runs generation of the mocks for go interfaces

The targets above must be run explicitly.

The `generate-mocks` target looks for `.mockery.yaml` files in the repo and it runs the mockery mock generator on each `.mockery.yaml` file it finds. This has the nice effect of allowing `.mockery.yaml` files to be in either the root of the repo or in subdirectories, so the choice of placement of `.mockery.yaml` files is left to the developer.


.mockery.yaml
vjayaramrh, line 1: what is the reason for deleting the .mockery.yaml file at the root of the project, this is the default behavior expected of mockery when searching for configuration file.

lapentad:
/approve 
tested on wsl

.mockery.yaml
liamfallon: I added the `.mockery.yaml` back in, we can have the default values we want to use for all interface mocks in the repo, and then use the `.mockery.yaml` for the specific mock generations we want to do for specific packages.

I think this is the best approach because there could be very different and conflicting setups for the mocks in different packages. In this way we can standardize the general stuff and allow for variations on packages.

vjayaramrh: I have not played with multiple .mockery.yaml files in different directories, is this something that you already tested out? Thanks

efiacor: @vjayaramrh I have tested it incorporating this PR - https://github.com/nephio-project/nephio/pull/511 and running the generate-mocks from root works ok.
I think for this project in particular it makes more sense to have specific mocks for each pkg/module.

vjayaramrh: @efiacor Thanks for confirming

vjayaramrh (approved):
/approve
/lgtm

.mockery.yaml
vjayaramrh: In my opinion, Where possible, if the info to generate mocks can be captured in a centralized way in the root directory would help look in a central place and reduce confusion

PR #722, opened 2024-04-25
Check if linting works on clean PR

liamfallon:
Linting is failing on https://github.com/nephio-project/nephio/pull/597, this PR is a test PR to see if linting is working on the main branch with just a linter upgrade.

PR #597, opened 2024-04-23
Upgrade Nephio to Go 1.22

liamfallon:
**What type of PR is this?**

> Uncomment only one  /kind <> line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>  /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it:**

- Updated go version in go.mod files
- Updated go version in github workflows
- Updated base image to 1.22.2-bookworm in Dockerfiles
- Updated Go/Golang-CI/GoSec versions in makefiles
- Updated reflect.TypeOf() to reflect.TypeFor[]() where relevant

**Which issue(s) this PR fixes:**
Fixes [Issue 510](https://github.com/nephio-project/nephio/issues/510).

**Special notes for your reviewer:**

**Does this PR introduce a user-facing change?:**
no

PR #723, opened 2024-04-26
Bumping gotests image version to next build

radoslawc:
golangci-lint update to work with go 1.22 updated other components as well
This resolves failing tests in https://github.com/nephio-project/nephio/pull/597 and https://github.com/nephio-project/nephio/pull/722


liamfallon (commented):
/lgtm
/approve
Thanks again @radoslawc

PR #734, opened 2024-05-14
Manual CodeQL configuration to mitigate tide merging problems

radoslawc:
CodeQL default setup has been causing problems by not recognising golang, adding manual configuration.


radoslawc:
@tliron @henderiw this will have to be merged manually since 'old' codeql is hanging there, but the one with manual configuration has passed.

Issue #779, opened 2024-07-25
Update nephio to use new nephio porch modules


efiacor:
Nephio controllers are still using the GoogelContonerTools porch version.
This needs to be updated to point to our new porch codebase.

PR #781, opened 2024-07-25
Update porch dependencies to nephio porch modules

efiacor:
#779 

Main change here was to remove the use of the replace directive for porch in the go.mod files.
Update the import paths.
Minor code updates.
Update the other dependencies to satisfy the existing codebase.


radoslawc:
/override presubmit-nephio-lichen  
this is false positive and needs further investigation

kispaljr:
changes look good and unit tests are all green on my machine

/lgtm

henderiw:
I could not test but overall
/lgtm

Issue #462, opened 2023-12-15
High Package Approval Time 


arora-sagar:
While checking for R2, I realized that the package approval time is more than R1. Sometimes even after 15 mins the approval doesn't happen so I have to approve it manually. 

I tried it several times and it seems that the issue is reproducible. 

@johnbelamaric if needed I can add the logs of the porch controller.

johnbelamaric:
https://github.com/nephio-project/nephio/pull/452

liamfallon:
related to https://github.com/nephio-project/nephio/issues/398

liamfallon:
Triaged
Triage Comment: Urgent

liamfallon:
I have run multiple Nephio end to end tests locally, but I am not seeing this issue on Porch. The packages are approved in a reasonable amount of time seconds/small number of minutes.

However, the resources associated with the packages themselves can take a long time to come up (~10 minutes for cluster creation over capi). This is not a porch issue but an issue in the package itself.

Logging from e2e test to follow once the PR below containing better logging is merged.
https://github.com/nephio-project/test-infra/pull/267

arora-sagar:
> I have run multiple Nephio end to end tests locally, but I am not seeing this issue on Porch. The packages are approved in a reasonable amount of time seconds/small number of minutes.
> 
> However, the resources associated with the packages themselves can take a long time to come up (~10 minutes for cluster creation over capi). This is not a porch issue but an issue in the package itself.
> 
> Logging from e2e test to follow once the PR below containing better logging is merged. [nephio-project/test-infra#267](https://github.com/nephio-project/test-infra/pull/267)

Thank you for looking into this. I totally agree and I learned about it when I was working on Onesummit demo. The problem is mostly related to the package. About cluster creation, I can understand 10 mins. But the same is not the case when NF package has a [dependencies CR](https://github.com/nephio-project/api/blob/main/config/crd/bases/req.nephio.org_dependencies.yaml). So it can be coming from functions, interface-fn, nfdeploy-fn or dnn-fn. 

I hope logging is able to show us the reason.

liamfallon:
Below is the timestamps extracted from the log from [this](http://prow.nephio.io/view/gs/prow-nephio-sig-release/logs/e2e-daily-ubuntu-focal-free5gc/1799456466965368832) test run on Saturday:


[approve.csv](https://github.com/user-attachments/files/15776453/approve.csv)

liamfallon:
[...]

Issue #398, opened 2023-09-20
Approval controller improvements


johnbelamaric:

johnbelamaric:
cc @arora-sagar

johnbelamaric:
related: https://github.com/kptdev/kpt/pull/4040

johnbelamaric:
/retitle Approval controller improvements

In the long run, we could design a more robust approval controller that has ApprovalPolicy resources that we can attach to packages. However, for now, we can make some iterative improvements. As a quick hack, we could make `approval.nephio.org/policy` a comma-delimited list (or create multiple annotations?), with the following policies which can be each independently enabled or disabled:

1. `initial` - the current policy, which will propose & approve a PackageRevision if there is no published version of that pacakage in the repository.
2. `proposed` - a new policy which will auto-approve any PackageRevision in a Proposed state
3. `delete-proposed` - a new policy which will auto-delete any PackageRevision in a DeletionProposed state

See https://github.com/nephio-project/nephio/blob/main/controllers/pkg/reconcilers/approval/ for the code. This could be a nice, well contained piece of functionality for someone wanting to get a taste of controller development.

Another little tweak would be to send an event when we approve. Right now, we only send events for failures or when we are "waiting". It would be a better UX to see an approval event as well. That would be here: https://github.com/nephio-project/nephio/blob/main/controllers/pkg/reconcilers/approval/reconciler.go#L202

cc @adetalhouet

johnbelamaric:
Oh - for 3) this will allow a cascading delete from PVS, because when you delete the PVS, it deletes the PV resources, which in turn results in the PackageRevision going to DeletionProposed.

liamfallon:
Triaged
Triage Comment: This is to do with the conditions the approval controller is waiting for do not happen, so it's not actually an approval controller issue, related to https://github.com/nephio-project/nephio/issues/462

johnbelamaric:
Looking at https://github.com/nephio-project/nephio/pull/797, I think we need another policy that just auto-approves everything. It should be really easy to implement. Basically, instead of "initial", we "always" or "ready" which approves any thing that has met the readiness gates. That will allow our controllers to make multiple revisions and avoid some of the race conditions inherent in the "initial" policy use.

kispaljr:
Continuing the conversation from the closed #797:
[...]

PR #797, opened 2024-08-15
Improve approval reconciler timings

efiacor:
Change approval controller PR Get to hit the api directly instead of reading from local cache.
Adjust the reque duration to prevent race condition.

During debugging the approval delay issue reported [here](https://github.com/nephio-project/nephio/issues/462) it became apparent that the packagerev being fetched was a cached version which didn't get updated for quite some time.
To circumvent this, we are retrieving the PR using the apiReader interface which bypasses the local cache and hits the k8s api directly.

### DIFF
--- a/controllers/pkg/reconcilers/approval/reconciler.go
+++ b/controllers/pkg/reconcilers/approval/reconciler.go
@@ -47,7 +47,7 @@ const (
 	DelayAnnotationName          = "approval.nephio.org/delay"
 	PolicyAnnotationName         = "approval.nephio.org/policy"
 	InitialPolicyAnnotationValue = "initial"
-	RequeueDuration              = 10 * time.Second
+	RequeueDuration              = 15 * time.Second
 )
 
 func init() {
@@ -67,6 +67,7 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c i
 		return nil, fmt.Errorf("cannot initialize, expecting controllerConfig, got: %s", reflect.TypeOf(c).Name())
 	}
 
+	r.apiReader = mgr.GetAPIReader()
 	r.baseClient = mgr.GetClient()
 	r.porchClient = cfg.PorchClient
 	r.porchRESTClient = cfg.PorchRESTClient
@@ -80,6 +81,7 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c i
 
 // reconciler reconciles a NetworkInstance object
 type reconciler struct {
+	apiReader       client.Reader
 	baseClient      client.Client
 	porchClient     client.Client
 	porchRESTClient rest.Interface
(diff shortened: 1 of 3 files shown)


efiacor:
#462

controllers/pkg/reconcilers/approval/reconciler.go
```diff
@@ -80,6 +81,7 @@ func (r *reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, c i
 
 // reconciler reconciles a NetworkInstance object
 type reconciler struct {
+	apiReader       client.Reader
```
vjayaramrh, line 84: Can you provide additional info on introducing the `apiReader`? Thanks

efiacor: Added additional info to the initial comment above.

kushnaidu (approved):

controllers/pkg/reconcilers/approval/reconciler.go
```diff
@@ -47,7 +47,7 @@ const (
 	DelayAnnotationName          = "approval.nephio.org/delay"
 	PolicyAnnotationName         = "approval.nephio.org/policy"
 	InitialPolicyAnnotationValue = "initial"
-	RequeueDuration              = 10 * time.Second
+	RequeueDuration              = 15 * time.Second
```
arora-sagar, line 50: @efiacor What does it signify? Should this value be statically calculated?

arora-sagar: How do we know that 15 is the right value?

efiacor: How would you propose that we statically calculate it?

arora-sagar: I don't know, how it is working or in the first place what are the factors which govern it. 
You changed it to 15 instead of 10 so I thought you would know the reason why exactly this value.

johnbelamaric (approved):
These changes look fine to me. However, it is clear that the "initial" policy isn't really meeting our needs. I suggest we add another policy to approval controller.

See: https://github.com/nephio-project/nephio/issues/398 - I will add another comment there.

efiacor:
> These changes look fine to me. However, it is clear that the "initial" policy isn't really meeting our needs. I suggest we add another policy to approval controller.
> 
> See: #398 - I will add another comment there.

Perfect. I will add a new "always" policy in the next PR.

controllers/pkg/reconcilers/approval/reconciler.go
kispaljr: > " I noticed intermittent issues where the AMF especially would go from approved - v1, to being reprocessed and a v2 branch getting created"

Do you know if it was the approval controller that approved the packagerevision prematurely, or was it some other actor?

johnbelamaric: The only other actor would be the test script itself or a human. Tweaking timings is a losing battle, the overall flow should be robust to race conditions. So, I using a different approval policy makes more sense.

efiacor: Still doing some debugging to determine the cause/source but it seems the pv controller is triggering a new "clone" when it finds a diff in the KptFile.
https://github.com/nephio-project/porch/blob/main/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go#L827
This comes from the generic specializer controller I believe. 
Even with using an "always" policy I am seeing some odd behaviour. Will continue with more refinements later.

PR #806, opened 2024-09-02
Add new "always" approval policy

efiacor:
Update pkgrev.get to use api Reader to bypass cache
Additional logging
Update kyaml versions to sync with porch version


controllers/pkg/reconcilers/generic-specializer/reconciler.go
```diff
@@ -296,12 +298,12 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		log.Info("generic specializer root kptfile", "packageName", pr.Spec.PackageName, "repository", pr.Spec.RepositoryName, "kptfile", kptfile)
 
-		kptf := kptfilelibv1.KptFile{Kptfile: rl.Items.GetRootKptfile()}
-		pr.Status.Conditions = porchcondition.GetPorchConditions(kptf.GetConditions())
+		// kptf := kptfilelibv1.KptFile{Kptfile: rl.Items.GetRootKptfile()}
```
liamfallon, line 301: Should we remove these commeted lines?

liamfallon:
Looks good. I'll download it and tryto run the e2e tests locally.

controllers/pkg/reconcilers/approval/reconciler.go
```diff
@@ -123,17 +122,17 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	if !pvReady {
-		r.recorder.Event(pr, corev1.EventTypeNormal,
-			"NotApproved", "owning PackageVariant not Ready")
+		r.recorder.Eventf(pr, corev1.EventTypeNormal,
+			"NotApproved", "owning PackageVariant for %s not Ready", pr.Spec.PackageName)
```
johnbelamaric, line 126: Since this is already attached to `pr`, I don't think we need the name in the message. You see these events by looking at `kubectl describe packagerevision foo` , so we shouldn't need to repeat `foo` here.

controllers/pkg/reconcilers/approval/reconciler.go
```diff
@@ -123,17 +122,17 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		return ctrl.Result{RequeueAfter: r.requeueDuration}, nil
 	}
 
 	// All policies require readiness gates to be met, so if they
 	// are not, we are done for now.
 	if !porchconds.PackageRevisionIsReady(pr.Spec.ReadinessGates, pr.Status.Conditions) {
-		r.recorder.Event(pr, corev1.EventTypeNormal,
-			"NotApproved", "readiness gates not met")
+		r.recorder.Eventf(pr, corev1.EventTypeNormal,
+			"NotApproved", "readiness gates not met for %s, in repo %s", pr.Spec.PackageName, pr.Spec.RepositoryName)
```
johnbelamaric, line 135: ditto

controllers/pkg/reconcilers/approval/reconciler.go
```diff
@@ -159,7 +160,7 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	if !approve {
 		r.recorder.Eventf(pr, corev1.EventTypeNormal,
-			"NotApproved", "approval policy %q not met", policy)
+			"NotApproved", "approval policy %q not met for %s", policy, pr.Spec.PackageName)
```
johnbelamaric, line 163: ditto

controllers/pkg/reconcilers/approval/reconciler.go
```diff
@@ -210,7 +211,7 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			"Error", "error %s: %s", action, err.Error())
 	} else {
 		r.recorder.Eventf(pr, corev1.EventTypeNormal,
-			reason, "all approval policies met")
+			reason, "all approval policies met for %s: %s", pr.Spec.PackageName, reason)
```
johnbelamaric, line 214: ditto

controllers/pkg/reconcilers/generic-specializer/reconciler.go
```diff
@@ -120,8 +122,8 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	if !pvReady {
-		r.recorder.Event(pr, corev1.EventTypeNormal,
-			"Waiting", "owning PackageVariant not Ready")
+		r.recorder.Eventf(pr, corev1.EventTypeNormal,
+			"Waiting", "owning PackageVariant for %s not Ready", pr.Spec.PackageName)
```
johnbelamaric, line 126: ditto

johnbelamaric (approved):
Ok, looks good. I am approving, but I do think it's not necessary to include the additional details in the event messages; the event is already attached to the package revision. I am unsure of whether there are optimizations in the event handling that may be defeated by adding additional info in the message rather than a static message. I guess it's still static, *for that PackageRevision*, so maybe it's ok.

efiacor:
> Ok, looks good. I am approving, but I do think it's not necessary to include the additional details in the event messages; the event is already attached to the package revision. I am unsure of whether there are optimizations in the event handling that may be defeated by adding additional info in the message rather than a static message. I guess it's still static, _for that PackageRevision_, so maybe it's ok.

Hi @johnbelamaric thanks for reviewing. Ye I just added the additional info for logging/debugging purposes. If you think it doesn't belong in the event, then maybe we should add it to a logger somewhere.

controllers/pkg/reconcilers/approval/reconciler.go
johnbelamaric: > I added these purely for additional logging detail. I found it difficult to debug with just the pr as it's just a long hex appended to the pr repo.

Ahh, right. Yeah, this change makes sense to me, then. Keep these here.

PR #811, opened 2024-09-18
Always run all tests

radoslawc:
There were some build fails that slipped trough tests resulting in broken build. Enabling 'always_run' as those are not long-running, just to be on safe side.


liamfallon (commented):
/approve
/lgtm

PR #849, opened 2025-02-05
Nephio operator for focom to oran-provisioning requests

dek5troza:
Very basic implementation of FOCOM operator, which reconciles FOCOM provisioning requests into oran provisioning requests. This PR contains very basic functionality based on the POCs explored earlier, and with the review, more functionality will be added if needed.

Reconciling FOCOM provisioning requests into oran provisioning requests, will look for correct TemplateInfo referenced in the request. Current validation only checks if the referenced TemplateInfo exists, with given name and version. Template parameters are not validated yet, it will be done in the future. Once the referenced template is validated, reconcile will proceed to find OCloudId referenced in the request and make sure it exists. From the OCloudId it will read the secret referenced in the OCloud.

In this early stage, secret contains kubeconfig for the OCloud cluster, which will be used to build client to interact with the remote cluster. Using this client reconciler will produce Oran provisioning requests on the remote cluster.

Later, this is planned to be changed to support REST-like interface on the remote cluster, rather then using kubeconfig. Secret in this phase should be endpoint and bearer token,

Project contains few test samples, using testenv, and there is usual e2e test from the operator-sdk that can validate operator deployment. This code is at the POC level, and needs more work to be production ready.

ORAN provisioning request CRD is located inside oran-provisioning-crd folder. Focom CRDS are generated by operator-sdk/kubebuilder from go structs in the api folder. Generated yaml files are located in config/crd/bases.

Current code does not cater for more then one-shot operations, and patching of ORAN provisioning requests is not supported yet. Validation of the TemplateInfo parameters is not implemented yet, potentially this can be done by validating webhook, making them immutable for a start.


default-go-lint.mk
```diff
@@ -1,4 +1,4 @@
-#  Copyright 2023 The Nephio Authors.
+#  Copyright 2023-2025 The Nephio Authors.
```
liamfallon, line 1:
```suggestion
#  Copyright 2023,2025 The Nephio Authors.
```

operators/focom-operator/hack/boilerplate.go.txt
liamfallon, line 1: What does this file do? It seems to be only the copyright notice?

liamfallon (commented):
The code is well structured and easy to read. It would be very good to get a walkthrough of the code at a future SIG-2 meeting.

krm-functions/lib/nad/v1/nad.go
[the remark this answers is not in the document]
efiacor: Agreed. Came up as a result of the golint upgrade.

operators/focom-operator/hack/boilerplate.go.txt
liamfallon: Hi @dek5troza thanks, that's great.

operators/focom-operator/api/focom/v1alpha1/focomprovisioningrequest_types.go
arora-sagar, line 1: @efiacor This whole folder should be in nephio/api repository instead of being here. 

The other operators are using it like that, also the CRD should not be here

operators/focom-operator/oran-provisioning-crd/provisioning-request-crd.yaml
arora-sagar, line 1: This CRD should be fulled directly from API repository

operators/focom-operator/test-data/red-hat-cluster-template.yaml
arora-sagar, line 1: Is this file used anywhere?

efiacor: I think it's used in the helpers_test ~L136

https://github.com/nephio-project/nephio/pull/849/files/59cf850cfa02db48d98de112f080e72e871aee99#diff-f0943a469b6d0bce80391a1d1f26f9d1d560018c72487aaae55c45516b99f004R136

operators/focom-operator/oran-provisioning-crd/provisioning-request-crd.yaml
efiacor: Ok. Workaround added to the helper_test setup and cleanup.

efiacor: https://github.com/nephio-project/nephio/pull/849/files#diff-f0943a469b6d0bce80391a1d1f26f9d1d560018c72487aaae55c45516b99f004R51

operators/focom-operator/api/focom/v1alpha1/focomprovisioningrequest_types.go
dek5troza: This is my understanding as well. CRD is owned by the focom-operator. It can be placed/copied elsewhere if needed, but the CRD itself (given this was built using operator-sdk) is generated from the go code. The entire codebase is self-contained project - I pointed this out, its structure is not the same as the operators you have already, which were built form "scratch".

operators/focom-operator/oran-provisioning-crd/provisioning-request-crd.yaml
dek5troza: This CRD can go elsewhere, its used only as an example and for testing, while the I wrote the code.

operators/focom-operator/test-data/red-hat-cluster-template.yaml
dek5troza: It's not used atm, intent was to use it in the test, instead of crafted cr.

arora-sagar: Then for this MR we can remove this file if it is not needed

operators/focom-operator/api/focom/v1alpha1/focomprovisioningrequest_types.go
arora-sagar: The other operators have used kubebuilder and they also use auto-generated files. 

Quite some files here are autogenerated --> https://github.com/nephio-project/api. 

Currently, all the controllers of Nephio reside in this repo, their CRD in the API repository, and their packaging in the catalog repository. I understand that this is not the most optimal way of placing the components, but to be fair with the whole project structuring we can follow the same approach for this MR.

krm-functions/lib/nad/v1/nad.go
arora-sagar, line 1: How are you planning to use the nad here? Is it needed at this moment?

efiacor: These required fixes were flagged by linting job - https://storage.googleapis.com/prow-nephio-sig-release/pr-logs/pull/nephio-project_nephio/849/presubmit-nephio-golangci-lint/1887251020145233920/build-log.txt

operators/focom-operator/test-data/red-hat-cluster-template.yaml
efiacor: removing

operators/focom-operator/oran-provisioning-crd/provisioning-request-crd.yaml
efiacor: resolved in previous commit

operators/focom-operator/api/focom/v1alpha1/focomprovisioningrequest_types.go
liamfallon: As a rule, generated files should not ever be checked in. They should be generated as part of the build as a precompile step in the build and placed in a temporary location were the compile step of the build can access them. This:
- reduces the risk of misalignment happening here the metadata used to generate the generated files gets out of sync with the compiled code
- reduces the risk of tooling used to generate code changing and causing different versions of generated code to be rendered for the same metadata
- reduces the volume of code that needs to be maintained
- avoids the risk of developers accidentally (or worse pragmatically) modifying the generated code

However, for R4 let's continue with the current approach and let's fix this in R5. I'll open an issue on this.

efiacor: I guess we can push the focom CRD to the api repo but not sure it's necessary. The current CRDs are getting picked up here - https://docs.nephio.org/docs/apis/topology-and-networking/
Do the ORAN usecase CRDs fall under Nephio topology and networking?

For the catalog pkg, the plan is to have a blueprint pkg using this one - https://github.com/nephio-project/nephio/pull/849/files#diff-ca25062a9c6702923f6312f24a4f0589ec31700cb74a6fc39b833321efe33d52
My proposal would be that we need to setup some additional build jobs, post R4, to push the artifacts (CRD(s) + kpt pkg) to their respective repos. For now, we will push the manually.

liamfallon: See https://github.com/nephio-project/nephio/issues/851

liamfallon: > Do the ORAN usecase CRDs fall under Nephio topology and networking?



> I guess we can push the focom CRD to the api repo but not sure it's necessary. The current CRDs are getting picked up here - https://docs.nephio.org/docs/apis/topology-and-networking/ Do the ORAN usecase CRDs fall under Nephio topology and networking?
> 
> For the catalog pkg, the plan is to have a blueprint pkg using this one - https://github.com/nephio-project/nephio/pull/849/files#diff-ca25062a9c6702923f6312f24a4f0589ec31700cb74a6fc39b833321efe33d52 My proposal would be that we need to setup some additional build jobs, post R4, to push the artifacts (CRD(s) + kpt pkg) to their respective repos. For now, we will push the manually.

I think it's better to have them int he API repo. Isn't the IMS one in there?

efiacor: It is. Maybe my naive understanding is incorrect, but does it belong to [Nephio Topology and Networking?](https://docs.nephio.org/docs/apis/topology-and-networking/)

efiacor: Will be addressed in R5
See - https://github.com/nephio-project/nephio/issues/852

default-go-test.mk
```diff
@@ -28,7 +26,7 @@ unit: test
 .PHONY: test
 test: ## Run unit tests (go test)
 ifeq ($(CONTAINER_RUNNABLE), 0)
-		$(RUN_CONTAINER_COMMAND) docker.io/library/golang:${GO_VERSION}-alpine3.19 \
+		$(RUN_CONTAINER_COMMAND) docker.io/nephio/gotests:1885274380137664512 \
```
vjayaramrh, line 29: What is the rationale not to use a generally available container image? Thanks

default-gosec.mk
```diff
@@ -12,15 +12,15 @@
...
-GOSEC_VER ?= 2.15.0
 GIT_ROOT_DIR ?= $(dir $(lastword $(MAKEFILE_LIST)))
 include $(GIT_ROOT_DIR)/detect-container-runtime.mk
 
 # Install link at https://github.com/securego/gosec#install if not running inside a container
 .PHONY: gosec
 gosec: ## Inspect the source code for security problems by scanning the Go Abstract Syntax Tree
 ifeq ($(CONTAINER_RUNNABLE), 0)
-		$(RUN_CONTAINER_COMMAND) docker.io/securego/gosec:${GOSEC_VER} ./...
+		$(RUN_CONTAINER_COMMAND) docker.io/nephio/gotests:1885274380137664512 gosec -fmt=html -out=gosec-results.html \
```
vjayaramrh, line 22: why not use the regular available gosec image? Thanks

default-go-test.mk
efiacor: Plan is to move the presubmits for here and porch to use this common image coming form the test-infra project.
https://github.com/nephio-project/test-infra/blob/main/images/gotests/Dockerfile

default-gosec.mk
efiacor: Plan is to move the presubmits for here and porch to use this common image coming form the test-infra project.
https://github.com/nephio-project/test-infra/blob/main/images/gotests/Dockerfile

arora-sagar:
I am fine with this MR, I already tested it end-to-end with O2IMS operator. It works fine. 

I am not a big fan of how it consumes the kubeconfig as a secret but that can be discussed later. 

/approve

default-gosec.mk
vjayaramrh: @efiacor Appreciate the response, 
From what I understand regarding container best practices, you want a container to contain only one application.

PR #864, opened 2025-02-20
Address ossf vuln

efiacor:
Addresses some of these issues:
https://github.com/nephio-project/nephio/security/code-scanning


krm-functions/configinject-fn/Dockerfile
```diff
@@ -12,14 +12,14 @@
...
+FROM golang:1.23.5-alpine@sha256:sha256:47d337594bd9e667d35514b241569f95fb6d95727c24b19468813d596d5ae596
 ENV CGO_ENABLED=0
 WORKDIR /go/src/
 COPY krm-functions/ krm-functions/
 WORKDIR /go/src/krm-functions/configinject-fn
 RUN go install
 RUN go build -o /usr/local/bin/function ./
 
-FROM gcr.io/distroless/static:latest
+FROM gcr.io/distroless/static:latest@sha256:3f2b64ef97bd285e36132c684e6b2ae8f2723293d09aae046196cca64251acac
```
Catalin-Stratulat-Ericsson, line 23: i could be wrong but when latest changes for this image will it not be a different sha? the sha is the hash of the exact commit/build right so its stagnant whereas latest changes over time? so can latest not become a different sha and therefore cause a problem later down the line?

Catalin-Stratulat-Ericsson: its probably fine for static versions e.g. golang:1.23.5-alpine my question regards images with the latest tag

krm-functions/gen-kustomize-res/Dockerfile
```diff
@@ -1,4 +1,4 @@
-#  Copyright 2025 The Nephio Authors.
+#  Copyright 2025-2025 The Nephio Authors.
```
Catalin-Stratulat-Ericsson, line 1: perhaps the script used to update the copyright should include a check for if the beginning year is the same as the latest year :D. to avoid this

Catalin-Stratulat-Ericsson (approved):
approved with comments

krm-functions/gen-kustomize-res/Dockerfile
efiacor: Done. Good catch

krm-functions/configinject-fn/Dockerfile
efiacor: Good catch. Corrected all distroless base images

PR #884, opened 2025-03-20
Refine pre and post submit job runs

efiacor:
Change prow jobs to only run if changed on:

run_if_changed: "(\\.go|Makefile|\\.mk)$"


liamfallon (commented):
/approve
/lgtm

Issue #808, opened 2024-09-17
Include a misconfigurations detector in Nephio


VedRatan:
## Checkov

Checkov is a powerful open-source static analysis tool designed for identifying misconfigurations and security vulnerabilities in infrastructure-as-code (IaC) templates. It supports various IaC frameworks like Terraform, Kubernetes, and CloudFormation, allowing developers to catch potential issues early in the development cycle.

## Importance of detecting misconfigurations
Detecting misconfigurations and security concerns with Checkov is crucial because even minor errors in configuration can lead to significant security risks, such as unauthorized access or data breaches. By using Checkov, we can automate the detection of these issues, ensuring that infrastructure is secure and compliant with best practices before deployment. This proactive approach helps in reducing the attack surface, improving overall security posture, and saving time and resources by addressing vulnerabilities early in the development process.

## Below are the findings after scanning Nephio cluster

[checkov-cluster.txt](https://github.com/user-attachments/files/17025211/checkov-cluster.txt)

## Below are the findings after scanning  a manifest file 

[checkov-file.txt](https://github.com/user-attachments/files/17025226/checkov-file.txt)

Here we see that there are multiple configurations best-practices and security-best practices that resources are failing currently. There are some default common misconfiguration and security detection policies pre-defined in checkov which are being applied against each of the resources present in the cluster.  Same is the case for pre-scan (scanning manifest only). 

We can include checkov to prevent any security misconfigurations and prevent the workloads to behave incorrectly beforehand by detecting these misconfigurations, and fixing them beforehand, here **checkov** can help us, as it also gives us the recommendations to mitigate the misconfigurations and security issues.

nyrahul:
Hey @VedRatan , you had done a prioritization task for checkov and shared an xls. We couldn't find it. Do you have it? Thanks

nyrahul:
CC: @gauravkr19

PR #919, opened 2025-05-19
add github action workflow and a folder containing script and other f…

gauravkr19:
This PR is for [issue-808](https://github.com/nephio-project/nephio/issues/808).


.github/checkov-scan/main.go
nyrahul, line 1: I am not sure if this file should be kept here. Ideally, this file should go into another repo and only the cli-tool generated should be used in other repos.

.github/workflows/checkov-scan.yml
```diff
@@ -0,0 +1,74 @@
...
+          echo "Using SKIP_PATHS: $SKIP_PATHS"
+          checkov -d . $SKIP_PATHS \
+          --skip-path ./.github/checkov-scan/checkov_raw_report.json \
+          --output json > ./.github/checkov-scan/checkov_raw_report.json || true
+
+      ### Step 3: Build and Run Go code to enrich severity
+      - name: _Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.24'
```
nyrahul, line 45: This should not be required if we remove the checkov-tool-utility code from this repo.

nyrahul (changes requested):
Many thanks for the PR. Please check the review comments

.github/checkov-scan/severity_map.json
efiacor, line 1: Similarly to @nyrahul s comment, I don't think we should have this go mod here.
Separate to that, I would strongly discourage having a customised setup like this. This would expire very quickly and would require ongoing maintenance. For scans like this I would purely rely on external tools and avoid maintaining our own.

.github/checkov-scan/go.mod
efiacor, line 1: This will trigger multiple make file targets to execute against this module, which will fail.

.github/checkov-scan/severity_map.json
gauravkr19: By default, Checkov will cause the CI to fail even on low-severity findings. However, the Checkov CLI supports skipping specific check IDs, which we can use to bypass low-severity issues.

This approach would require maintaining a list of low-severity check IDs to be excluded during the scan.

Let me know if this approach is acceptable—if so, I’ll go ahead and update the PR accordingly.

nyrahul: Got it. I will bring this to the notice of SIG-Security and see if we could align on prioritizing porch. I see the point. Thanks

.github/checkov-scan/go.mod
gauravkr19: Hi @efiacor - Should we implement the Checkov scan with soft-fail. It will publish the report but not fail any PR. Existing issues can be baselined.

efiacor: Hi @gauravkr19 .
I think that might be a better option for now. The utility tool would require maintenance which we unfortunately don't have resources for. We could move the tool itself to the test-infra repo somewhere if you would like to keep it. Somewhere here maybe - https://github.com/nephio-project/test-infra/tree/main/images

PR #944, opened 2025-06-26
Rename kpt-argo plugins

efiacor:
To align with https://github.com/nephio-project/test-infra/pull/352/files


dkosteck (approved):
Looks good, but let's figure out if we want latest or another image tag before merge

PR #1082, opened 2026-03-29
fix: resolve concurrent state corruption in NetworkController

pulkitvats2007-crypto:
###  Bug: Concurrent State Corruption in NetworkController

The `NetworkController` had a critical concurrency flaw where per-reconciliation state was stored on the reconciler struct itself. Since the reconciler is a shared singleton, concurrent reconciliation loops ended up mutating the same `resources` (and previously `devices`) fields.

In a real-world Nephio setup—where multiple `Network` CRs are reconciled simultaneously—this caused state from one reconciliation to leak into another. As a result, configurations such as IPAM allocations, VLAN assignments, or OpenConfig data intended for one network could be incorrectly applied to a different one.

This issue was particularly dangerous because it often **failed silently**: the controller would report a successful reconciliation while leaving the underlying infrastructure in an inconsistent or incorrect state. In some cases, concurrent writes to shared fields could also trigger race conditions and lead to controller panics.

---

###  Fix: Isolate Reconciliation State & Ensure Thread Safety

This PR resolves the issue by eliminating shared mutable state from the reconciler and ensuring that each reconciliation loop operates on its own isolated data.

Key changes:

- **Removed shared fields**: Eliminated `resources` and unused `devices` from the reconciler struct to prevent cross-request mutation.
- **Scoped state locally**: Moved initialization of the `resources` object into the `Reconcile` function so each invocation gets a fresh instance.
- **Explicit dependency passing**: Updated helper functions like `applyInitialResources` and `getNewResources` to accept `resources` as a parameter instead of relying on struct-level state.
- **Cleaned up dead code**: Removed the unused `devices` field to simplify the controller and reduce confusion.

---

###  Result

The `NetworkController` is now thread-safe and behaves correctly under concurrent reconciliation scenarios. Each `Network` CR is processed independently, eliminating cross-resource contamination and ensuring reliable, predictable infrastructure configuration—especially important in Nephio’s GitOps-driven workflows where batch updates are common.

### DIFF
--- a/controllers/pkg/go.mod
+++ b/controllers/pkg/go.mod
@@ -13,7 +13,6 @@ require (
 	code.gitea.io/sdk/gitea v0.22.1
 	github.com/go-logr/logr v1.4.3
 	github.com/google/go-cmp v0.7.0
-	github.com/henderiw-nephio/network v0.0.0-20231206051529-4287dc43f8a6
 	github.com/kptdev/kpt v1.0.0-beta.60
 	github.com/kptdev/krm-functions-sdk/go/fn v1.0.1
 	github.com/nephio-project/api v1.0.1-0.20250218114915-854faaf69fd0 //v4.0.0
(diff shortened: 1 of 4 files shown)


liamfallon (commented):
Why do we need this dependency?

require github.com/henderiw-nephio/network v0.0.0-20231206051529-4287dc43f8a6

liamfallon (approved):

liamfallon:
Please fix the merge conflicts.

liamfallon:
Please resolve the merge conflict.

PR #1094, opened 2026-04-15
fix: sync KRM function resource deletions in specializers

pulkitvats2007-crypto:
## Summary

Fixes a critical issue where KRM function pipelines were unable to delete resources due to append-only reconciliation logic in the Specializer controllers.

Previously, controllers (`generic-specializer`, `ipam-specializer`, `vlan-specializer`) rebuilt `prr.Spec.Resources` by iterating only over `rl.Items`. Since deleted resources are absent from `rl.Items`, they were never removed from the map, resulting in stale ("ghost") resources persisting in the package.

This PR introduces explicit deletion synchronization:
- Capture original resource paths before running the KRM pipeline.
- Capture surviving paths after pipeline execution.
- Diff both sets and explicitly remove missing entries using `delete()`.

This ensures full state reconciliation (create/update/delete) instead of append-only behavior.

---

## Bug

**Root Cause:**  
The controller assumed that iterating over the mutated `rl.Items` was sufficient to sync state back to `prr.Spec.Resources`. This ignored deletion semantics entirely.

**Resulting Issue:**  
- Resources deleted by KRM functions (e.g., via `condkptsdk` GC) were silently retained.
- No errors or signals were emitted, leading to misleading "successful" reconciliations.

---

## Fix

- Added path tracking (`originalPaths`, `survivingPaths`)
- Implemented `syncDeletedResources` helper to reconcile deletions
- Applied fix consistently across:
  - `generic-specializer`
  - `ipam-specializer`
  - `vlan-specializer`
- Added unit tests to prevent regression

---

## Impact

- Eliminates ghost resources from PackageRevisions
- Restores correct KRM deletion semantics
- Prevents downstream GitOps drift and infra conflicts (e.g., duplicate IPs)
- Improves reliability of Nephio automation in dynamic/topology-changing environments

### DIFF
--- a/controllers/pkg/reconcilers/generic-specializer/reconciler.go
+++ b/controllers/pkg/reconcilers/generic-specializer/reconciler.go
@@ -161,6 +161,13 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			return ctrl.Result{}, errors.Wrap(err, "cannot get resourceList")
 		}
 
+		originalPaths := map[string]struct{}{}
+		for _, o := range rl.Items {
+			if path := o.GetAnnotation(kioutil.PathAnnotation); path != "" {
+				originalPaths[path] = struct{}{}
+			}
+		}
+
 		if porchcondition.HasSpecificTypeConditions(pr.Status.Conditions, kptfilelibv1.GetConditionType(&ipamFor)) {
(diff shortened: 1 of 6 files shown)


liamfallon (commented):
I don't have time at the moment to revie this PR.

liamfallon (approved):

PR #1105, opened 2026-04-20
generic-specializer: skip condition walk when Kptfile.Status is nil

### LINKED ISSUE #909: Nil Pointer Dereference in Generic Specializer when Processing Kptfile without Status
The generic-specializer processes the `Kptfile` of a package revision by parsing it into a Go struct and directly accessing its `Status` field:

```go
kptfile, err := kptf.GetGoStruct()
if err != nil {
	log.Error(err, "cannot get gostruct from kubeobject")
	continue
}
for _, c := range kptfile.Status.Conditions {
	// ...
}
```

At runtime, if the `Kptfile` does not contain a `status` block (which is common for newly cloned repositories, raw drafts, or manually authored packages), `kptfile.Status` will be `nil` because `kptv1.KptFile` defines `Status` as an optional pointer (`*Status`). Attempting to iterate over `kptfile.Status.Conditions` triggers a `panic: runtime error: invalid memory address or nil pointer dereference`. 

This permanently halts the `PackageRevision` lifecycle because the panic is caught by the `controller-runtime` worker but the reconciliation is unable to progress.

I have submitted a PR that adds a defensive `nil` check before accessing `kptfile.Status.Conditions`.

### LINKED ISSUE #1099: Nil Pointer Dereference in Generic Specializer when Processing Kptfile without Status
The generic-specializer processes the `Kptfile` of a package revision by parsing it into a Go struct and directly accessing its `Status` field:

```go
kptfile, err := kptf.GetGoStruct()
if err != nil {
	log.Error(err, "cannot get gostruct from kubeobject")
	continue
}
for _, c := range kptfile.Status.Conditions {
	// ...
}
```

At runtime, if the `Kptfile` does not contain a `status` block (which is common for newly cloned repositories, raw drafts, or manually authored packages), `kptfile.Status` will be `nil` because `kptv1.KptFile` defines `Status` as an optional pointer (`*Status`). Attempting to iterate over `kptfile.Status.Conditions` triggers a `panic: runtime error: invalid memory address or nil pointer dereference`. 

This permanently halts the `PackageRevision` lifecycle because the panic is caught by the `controller-runtime` worker but the reconciliation is unable to progress.

I have submitted a PR that adds a defensive `nil` check before accessing `kptfile.Status.Conditions`.

SAY-5:
## Summary

`kptv1.KptFile.Status` is defined as `*Status` (optional), so newly cloned repos, raw drafts, and hand-authored packages reach the generic-specializer reconciler with `Status == nil`. The reconciler ranged over `kptfile.Status.Conditions` and then called `porchcondition.GetPorchConditions(kptfile.Status.Conditions)` without checking, so these common shapes panicked and permanently stalled the `PackageRevision` lifecycle (#1099).

Fixes nephio-project/porch#909.

## Change

`controllers/pkg/reconcilers/generic-specializer/reconciler.go`, wrap the condition walk + readiness-gate check in `if kptfile.Status != nil { ... }`. A Kptfile with no status has no conditions to walk anyway, so the ready-gate check on this path is trivially false in that case.

## Testing

- `go build ./controllers/...` passes.
- Manual: fed the reconciler a `PackageRevision` whose Kptfile has no `status` block. Previously it panicked inside the generic-specializer goroutine; now it logs nothing and moves on, and subsequent revisions for the same package progress normally.

### DIFF
--- a/controllers/pkg/reconcilers/generic-specializer/reconciler.go
+++ b/controllers/pkg/reconcilers/generic-specializer/reconciler.go
@@ -279,16 +279,23 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 					log.Error(err, "cannot get gostruct from kubeobject")
 					continue
 				}
-				for _, c := range kptfile.Status.Conditions {
-					if strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&vlanFor)+".") ||
-						strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&ipamFor)+".") ||
-						strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&configInjectFor)+".") {
-						log.Info("generic specializer conditions", "packageName", pr.Spec.PackageName, "repository", pr.Spec.RepositoryName, "status", c.Status, "condition", c.Type, "message", c.Message)
+				// kptv1.KptFile.Status is an optional *Status; newly
+				// cloned repos, raw drafts, and hand-authored packages
+				// reach the reconciler with Status == nil. Ranging over
+				// a nil Conditions slice below used to panic and stall
+				// the whole PackageRevision lifecycle (#1099).
+				if kptfile.Status != nil {
+					for _, c := range kptfile.Status.Conditions {
+						if strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&vlanFor)+".") ||
+							strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&ipamFor)+".") ||
+							strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&configInjectFor)+".") {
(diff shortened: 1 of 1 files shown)


controllers/pkg/reconcilers/generic-specializer/reconciler.go
```diff
@@ -279,16 +279,23 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 					log.Error(err, "cannot get gostruct from kubeobject")
 					continue
 				}
-				for _, c := range kptfile.Status.Conditions {
-					if strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&vlanFor)+".") ||
-						strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&ipamFor)+".") ||
-						strings.HasPrefix(c.Type, kptfilelibv1.GetConditionType(&configInjectFor)+".") {
-						log.Info("generic specializer conditions", "packageName", pr.Spec.PackageName, "repository", pr.Spec.RepositoryName, "status", c.Status, "condition", c.Type, "message", c.Message)
+				// kptv1.KptFile.Status is an optional *Status; newly
```
liamfallon, line 282: Can `KptFile.Status` ever be `nil`?

PR #1168, opened 2026-08-16
[security][focom] Verify Kubernetes API TLS certificates by default

thc1006:
The FOCOM NBI Porch client decides whether to verify the Kubernetes API server certificate from `os.Getenv("PORCH_HTTPS_VERIFY") == "true"`, and `newPorchStorageWithToken` feeds the negation of that into `tls.Config.InsecureSkipVerify`. Anything that is not the exact string `true` leaves verification off, the variable being unset included. Both shipped manifests set it to `"false"`, so a deployment built from `config/` or from the kpt package attaches its service account bearer token to every Porch request over a connection whose peer was never authenticated. CWE-295.

This makes verification the default, moves the TLS setup onto client-go, and renames the opt-out so it cannot be reached by omission.

## The bundle was already mounted

The Kubernetes documentation on [accessing the API from a Pod](https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/):

> If available, a certificate bundle is placed into the filesystem tree of each container at `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, and should be used to verify the serving certificate of the API server.

That bundle is already mounted in the FOCOM pod. It was never read, while the token sitting next to it was attached to every request by `makeRequest`. `cmd/main.go` grants that service account `create;delete;get;list;patch;update;watch` on `porch.kpt.dev` packagerevisions and `get;list;watch` on secrets, which is what the token is worth to whoever intercepts it: read what the operator reads, and rewrite the package revisions GitOps then applies to the cluster.

The same page also says which address the certificate is good for:

> Kubernetes does not guarantee that the API server has a valid certificate for the hostname `kubernetes.default.svc`; however, the control plane **is** expected to present a valid certificate for the hostname or IP address that `$KUBERNETES_SERVICE_HOST` represents.

Both manifests pinned `KUBERNETES_BASE_URL=https://kubernetes.default.svc`, which is the address without the guarantee. Turning verification on while keeping that value would trade a silent failure for a noisy one on any cluster whose serving certificate lacks that SAN, so the variable goes from the manifests and the address comes from `KUBERNETES_SERVICE_HOST`/`PORT` in a pod. It stays available as an override for anyone reaching the API through a proxy.

Unlike the o2ims operator, nothing in another repository has to be published in step with this one.

An installed Deployment is a different question, and I had this wrong. Editing the manifests in git does not edit a cluster. A deployment applied before this still carries

```
KUBERNETES_BASE_URL=https://kubernetes.default.svc
PORCH_HTTPS_VERIFY=false
```

and if the image is upgraded without re-applying the package, that explicit URL still wins over the address the pod is told about. `PORCH_HTTPS_VERIFY` goes inert either way, so verification does turn on, but against the address without the SAN guarantee. A fresh install needs no configuration; an upgrade of an existing one needs those two variables removed. Anyone who set `KUBERNETES_BASE_URL` deliberately, to reach the API through a proxy, keeps it and supplies that endpoint's CA.

## What the client looks like now

The hand-built `http.Transport` is gone and connection settings come from a `rest.Config`. `rest.HTTPClientFor` turns that into the HTTP client, which is the path `newPorchStorageFromKubeconfig` already used, so the two constructors now differ only in where their credentials come from. Outside a cluster a CA bundle can be supplied through `KUBERNETES_CA_FILE` or `PorchStorageConfig.CAFile`, and an unreadable bundle fails at construction rather than degrading to an unverified connection.

Host and CA are read from the environment rather than through `rest.InClusterConfig`. That function reads the fixed service account token before it will return either, so a pod running with `automountServiceAccountToken: false` and its token projected somewhere else, which Kubernetes supports, supplied everything needed and was still refused. Reading the two variables directly also removes the earlier choice between swallowing every `InClusterConfig` error and failing on one that never mattered: half a service environment is a configuration error, and the in-pod CA requirement is enforced on its own.

The bundle follows the address rather than the pod:

```go
	origin := apiServerOrigin(restConfig.Host)
	if caFile == "" && inClusterHost != "" &&
		(origin == apiServerOrigin(inClusterHost) ||
			origin == apiServerOrigin(defaultKubernetesURL)) {
		caFile = inClusterCAFile // required, not best effort
	}
```

The first draft applied the cluster CA whenever `KUBERNETES_SERVICE_HOST` was set. An operator in a pod pointed at some other Porch endpoint then had the cluster CA pinned to a server that CA does not certify, and every connection to it would fail. The cluster serves its API at the advertised address and at `https://kubernetes.default.svc`; anything else keeps the system roots, so the certificate is still checked, against the right authorities.

That comparison is by endpoint and not by string. `https://10.96.0.1` and `https://10.96.0.1:443` are one endpoint with one certificate, as are the short and long spellings of an IPv6 address and any casing of a DNS name. Comparing URLs as written sent four such forms to the system roots:

```console
https://10.96.0.1:443                  -> cluster CA
https://10.96.0.1                      -> public roots
https://kubernetes.default.svc         -> cluster CA
https://kubernetes.default.svc:443     -> public roots
```

That fails closed on a private-CA cluster rather than weakening anything, but it fails a configuration that is correct. `apiServerOrigin` normalises scheme, host and effective port first.

Fuzzing the validator against what the transport actually dials turned up one class `url.Parse` does not cover. It refuses a percent escape in the host, but keeps a non-ASCII name as written, and `net/http` maps it before dialling:

```console
"https://ⓔxample.com"    parsed host "ⓔxample.com"
                         dialled     "example.com"
```

So an operator can name one host, have its origin computed from that name, and reach another. The host has to be a shape every parser reads the same way now: an IP literal, or a DNS name of letter-digit-hyphen labels. Of two hundred thousand random authorities drawn from an alphabet of separators, escapes and full-width lookalikes, seventeen hundred are accepted and every one parses to a host already in that shape, which is the shape `idna.ToASCII` leaves alone. Port zero is refused too, since `url.Parse` only checks that a port is digits. o2ims arrived at the same rule in #1170, and the two now return the same verdict for every address in a shared corpus of twenty, valid and hostile alike.

Three other paths that could have broken the invariant were closed after review. The endpoint must parse as an absolute `https` URL with a host and no embedded credentials, because a plaintext `KUBERNETES_BASE_URL` used to be accepted verbatim with `makeRequest` sending the token over it and no opt-out required; the skip flag deliberately does not make plaintext legal. In a pod with no readable CA the connection fell back to the system roots, which is a wider trust boundary rather than a failure, so that now refuses to start unless a bundle is supplied or verification is explicitly skipped. And a CA bundle combined with the skip flag is refused rather than quietly discarded, with the flag clearing only the bundle instead of the whole TLS config, so `ServerName` and client certificates survive it.

The opt-out is `UNSAFE_SKIP_TLS_VERIFY`, parsed with `strconv.ParseBool` so a typo cannot relax it, and logged at startup when engaged. It is named for what it does rather than as a reversed `_VERIFY` flag, because a flag that disables verification by being absent is how this default became insecure. `PORCH_HTTPS_VERIFY` is not read at all any more; honouring it would preserve exactly the value the manifests encode. Startup says so if it is still set.

## Proof manifests

The rendered deployment is the easiest place to see it. On `main`, `kustomize build config/default` ships the insecure setting as a literal:

```console
        - name: PORCH_REPOSITORY
          value: focom-resources
        - name: KUBERNETES_BASE_URL
          value: https://kubernetes.default.svc
        - name: PORCH_HTTPS_VERIFY
          value: "false"
```

With this PR neither variable is rendered at all, so nothing has to opt back into verification:

```console
        - name: PORCH_REPOSITORY
          value: focom-resources
        image: docker.io/nephio/focom-operator:latest
```

Twenty one tests cover the boundary, against an `httptest.NewTLSServer` whose certificate no trust store knows about, the way a cluster CA is unknown to one:

```console
$ go test ./internal/nbi/storage/ -run TestPorchStorageTLS -v
--- PASS: TestPorchStorageTLS_VerifiedByDefault
--- PASS: TestPorchStorageTLS_TrustsConfiguredCA
--- PASS: TestPorchStorageTLS_CAFileFromEnv
--- PASS: TestPorchStorageTLS_UnreadableCAFile
--- PASS: TestPorchStorageTLS_RejectsPlainHTTP
--- PASS: TestPorchStorageTLS_RejectsUnusableURL
--- PASS: TestPorchStorageTLS_RejectsCAWithSkipVerify
--- PASS: TestPorchStorageTLS_CustomTokenInPod
--- PASS: TestPorchStorageTLS_InPodWithoutCAFailsClosed
--- PASS: TestPorchStorageTLS_InPodWithMalformedCAFailsClosed
--- PASS: TestPorchStorageTLS_OtherEndpointKeepsSystemRoots
--- PASS: TestPorchStorageTLS_ClusterNameUsesClusterCA
--- PASS: TestPorchStorageTLS_UnusableHostsAreRefused
--- PASS: TestPorchStorageTLS_APIServerOrigin
--- PASS: TestPorchStorageTLS_EmptyCAIsRefused
--- PASS: TestPorchStorageTLS_LeadingZeroPortStillNeedsTheClusterCA
--- PASS: TestPorchStorageTLS_RequestPathSurvivesTheBaseURL
--- PASS: TestPorchStorageTLS_UnusableEndpointIsRefusedBeforeTheCA
--- PASS: TestPorchStorageTLS_EquivalentClusterAddresses
--- PASS: TestPorchStorageTLS_PartialServiceEnvIsRefused
--- PASS: TestPorchStorageTLS_SkipVerifyOptIn
ok      github.com/nephio-project/nephio/operators/focom-operator/internal/nbi/storage
```

A passing test proves little on its own, so each one was run against the change it is supposed to catch. Reverting the secure default fails `VerifiedByDefault` and `UnreadableCAFile`. Removing the https check fails `RejectsPlainHTTP`. Removing the in-pod CA requirement fails `InPodWithoutCAFailsClosed`. Removing the partial-environment check fails `PartialServiceEnvIsRefused`. Unscoping the cluster CA fails `OtherEndpointKeepsSystemRoots`, and dropping the in-cluster name from that rule fails `ClusterNameUsesClusterCA`.

The suite also has to give the same answer wherever it runs. `TestMain` clears the variables the kubelet injects, because this code reads them now. Without it, running the package inside a pod picks up whichever cluster that is, and four pre-existing tests about token and kubeconfig resolution fail on a CA bundle they were never about:

```console
$ env KUBERNETES_SERVICE_HOST=10.96.0.1 KUBERNETES_SERVICE_PORT=443 \
      go test -count=1 -shuffle=on ./internal/nbi/storage/     # before TestMain
--- FAIL: TestNewPorchStorage_WithConfig
    open /var/run/secrets/kubernetes.io/serviceaccount/ca.crt: no such file or directory
... 6 failures

$ env KUBERNETES_SERVICE_HOST=10.96.0.1 KUBERNETES_SERVICE_PORT=443 \
      go test -count=1 -shuffle=on ./internal/nbi/storage/     # after
ok      github.com/nephio-project/nephio/operators/focom-operator/internal/nbi/storage
```

The two in-pod cases used to depend on the machine running them. `inClusterCAFile` was a constant, so a test could not hide the bundle Kubernetes mounts at that path in any pod carrying the service account credential, a presubmit pod included. Making it a variable that tests redirect fixes that, and with a bundle placed at the real path the difference shows:

```console
$ sudo cp /etc/ssl/certs/ca-certificates.crt /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
$ go test ./internal/nbi/storage/ -run 'TestPorchStorageTLS_InPod'      # with the fix
ok

$ # with the test unable to redirect the path, as before
--- FAIL: TestPorchStorageTLS_InPodWithoutCAFailsClosed
```

It also makes a malformed bundle at the fixed path testable for the first time, which an earlier revision of this description listed as out of reach.

None of that means anything unless CI runs it. `presubmit-nephio-go-test` runs `make unit`, whose storage step carries a `-run` allowlist, so a new test is invisible to CI unless it is named there. The shared `TestPorchStorageTLS` prefix is added to the allowlist, which covers future tests in the group without touching the Makefile again:

```console
$ make unit | grep -c '^=== RUN'
231                    # 186 before this change

$ make gosec
  Files  : 51
  Issues : 0

$ go test ./internal/... -count=1
ok      .../internal/controller           0.129s
ok      .../internal/nbi                  0.039s
ok      .../internal/nbi/models           0.020s
ok      .../internal/nbi/services         0.027s
ok      .../internal/nbi/storage          6.406s
ok      .../internal/nbi/validation       0.038s
```

`gosec` also needs one fewer `#nosec` to stay clean, since `G402` was suppressed at the line this deletes. `test/e2e` is not in that list because it fails identically on `main` here; it needs a live cluster and network access to install the Prometheus operator.

## One behaviour change worth knowing about

client-go installs `http.ProxyFromEnvironment` and the transport it replaces set no proxy at all. Measured rather than reasoned about:

```console
main (hand-built transport)   proxy not contacted
this branch (HTTPClientFor)   proxy CONTACTED: "CONNECT 10.96.0.1:443 HTTP/1.1"
```

So on a cluster that exports `HTTPS_PROXY` to workloads with no `NO_PROXY` entry for the API server, this connection starts going through it. Two things make that acceptable rather than a regression. The manager's own client in the same binary has always gone through client-go and therefore always honoured those variables, so this makes the Porch client consistent rather than novel. And pinning the cluster CA means a proxy that terminates TLS fails the handshake instead of quietly reading the token. `docs/PORCH_SETUP.md` says so now.

## Effect on existing deployments

In-cluster deployments need no configuration, since the CA bundle and the API address both come from the pod. Anyone running the operator outside a cluster against an endpoint named by `KUBERNETES_BASE_URL` now needs `KUBERNETES_CA_FILE`, or `useKubeconfig`, which carries its cluster CA already. `PORCH_HTTPS_VERIFY` becomes inert, and `UNSAFE_SKIP_TLS_VERIFY=true` restores the old behaviour for development, loudly. `PorchStorageConfig.HTTPSVerify` is replaced by `InsecureSkipTLSVerify` and `CAFile` so the zero value is the safe one, with `internal/nbi/testconfig.go` and its example following.

## Left out on purpose

Most of the storage package is outside the `make unit` allowlist, so it never runs in CI either. Widening that is a separate question, since some of those tests read `KUBECONFIG` and the ambient environment, which is presumably why the list exists at all.

`makeRequest` still sets the `Authorization` header by hand from a token read once at startup, so client-go's `BearerTokenFile` refresh goes unused and a rotated projected token will eventually 401, and an empty `Bearer ` suppresses exec auth under a kubeconfig. That is #1171. The fix is to put credentials on the `rest.Config` and reuse the manager's config instead of rediscovering one, and it reaches around thirty test constructions that build `PorchStorage` directly with a token field, which is why it is not here. The documentation claiming kubeconfig exec support has been corrected in this PR rather than left standing.

The o2ims operator had the same shape with a worse outcome, `bool(os.getenv("HTTPS_VERIFY", False))` being `False` when unset and `True` when set to the string `"false"`. That went in as #1170.

## Notes for maintainers

`operators/focom-operator/docs/{PORCH_SETUP,DEPLOYMENT}.md` are updated here. I searched nephio-project/docs for `PORCH_HTTPS_VERIFY`, `HTTPS_VERIFY` and the FOCOM install guide and found no reference to these variables, so a companion PR there does not look necessary. Happy to raise one if you disagree.

Release triage is yours to set. The change is confined to the FOCOM operator's Porch client and its manifests.

`SECURITY.md` asks that vulnerabilities not be reported through public issues, and I have not opened one. Say the word and I will route this through sig-security@lists.nephio.org before it goes further.

### DIFF
--- a/operators/focom-operator/Makefile
+++ b/operators/focom-operator/Makefile
@@ -141,7 +141,7 @@ unit-tests: fmt vet ## Run unit tests only (no external dependencies required).
 		./internal/nbi/models/ \
 		./internal/nbi/validation/
 	@go test -v -coverprofile=coverage-storage.out ./internal/nbi/storage/ \
-		-run "TestInMemory|TestCreateResourceYAML|TestParseResourceYAML|TestRoundTrip|TestExtractResourceID|TestGenerateWorkspaceName|TestSetResource|TestConvertTo|TestCreateKptfile"
+		-run "TestInMemory|TestCreateResourceYAML|TestParseResourceYAML|TestRoundTrip|TestExtractResourceID|TestGenerateWorkspaceName|TestSetResource|TestConvertTo|TestCreateKptfile|TestPorchStorageTLS"
 	@go test -v -coverprofile=coverage-controller.out \
 		./internal/controller/ -run TestValidateTemplateAlignment || echo "Controller tests skipped"
 	@echo ""
@@ -163,7 +163,7 @@ ci-tests: fmt vet ## Run the same tests as CI pipeline (unit tests only, fast ex
 		./internal/nbi/models/ \
 		./internal/nbi/validation/
 	@go test -v -coverprofile=coverage-storage.out ./internal/nbi/storage/ \
-		-run "TestInMemory|TestCreateResourceYAML|TestParseResourceYAML|TestRoundTrip|TestExtractResourceID|TestGenerateWorkspaceName|TestSetResource|TestConvertTo|TestCreateKptfile"
+		-run "TestInMemory|TestCreateResourceYAML|TestParseResourceYAML|TestRoundTrip|TestExtractResourceID|TestGenerateWorkspaceName|TestSetResource|TestConvertTo|TestCreateKptfile|TestPorchStorageTLS"
 	@go test -v -coverprofile=coverage-controller.out \
 		./internal/controller/ -run TestValidateTemplateAlignment || echo "Controller tests skipped"
 	@echo ""
--- a/operators/focom-operator/cmd/main.go
+++ b/operators/focom-operator/cmd/main.go
@@ -33,6 +33,7 @@ import (
 	"fmt"
 	focomv1alpha1 "github.com/nephio-project/nephio/operators/focom-operator/api/focom/v1alpha1"
 	"os"
+	"strconv"
 	"sync"
 	"time"
 
@@ -73,6 +74,13 @@ func init() {
 	// +kubebuilder:scaffold:scheme
 }
 
+// envBool reports whether the named environment variable holds a true value.
+// Unset, empty and unparseable all read as false.
+func envBool(name string) bool {
+	value, err := strconv.ParseBool(os.Getenv(name))
+	return err == nil && value
+}
+
 // initializeNBISystem initializes the complete NBI system with all components
 func initializeNBISystem(mgr ctrl.Manager, nbiConfig *config.NBIConfig) (*nbi.Server, error) {
 	setupLog.Info("Initializing NBI system", "stage", nbiConfig.Stage, "storageBackend", nbiConfig.StorageBackend)
@@ -96,15 +104,23 @@ func initializeNBISystem(mgr ctrl.Manager, nbiConfig *config.NBIConfig) (*nbi.Se
 			repository = "focom-resources"
 		}
 
-		httpsVerify := os.Getenv("PORCH_HTTPS_VERIFY") == "true"
+		// Certificates are verified unless development explicitly opts out.
+		insecureSkipTLSVerify := envBool("UNSAFE_SKIP_TLS_VERIFY")
+		if insecureSkipTLSVerify {
+			setupLog.Info("UNSAFE_SKIP_TLS_VERIFY is set: the Kubernetes API server certificate will NOT be verified " +
+				"and the service account token can be intercepted. Never enable this outside development.")
+		}
+		if _, ok := os.LookupEnv("PORCH_HTTPS_VERIFY"); ok {
+			setupLog.Info("PORCH_HTTPS_VERIFY is no longer supported and is ignored; certificates are always verified " +
+				"unless UNSAFE_SKIP_TLS_VERIFY=true")
+		}
 
 		porchConfig := &storage.PorchStorageConfig{
-			Namespace:   namespace,
-			Repository:  repository,
-			HTTPSVerify: httpsVerify,
-			// KubernetesURL and Token will be auto-detected from environment
-			// KUBERNETES_BASE_URL env var or default to "https://kubernetes.default.svc"
-			// TOKEN env var or service account token file
+			Namespace:             namespace,
+			Repository:            repository,
+			InsecureSkipTLSVerify: insecureSkipTLSVerify,
+			// KubernetesURL, Token and the API server CA are auto-detected from
+			// the environment: KUBERNETES_BASE_URL, TOKEN, KUBERNETES_CA_FILE
 		}
 
 		var err error
--- a/operators/focom-operator/config/manager/manager.yaml
+++ b/operators/focom-operator/config/manager/manager.yaml
@@ -78,10 +78,6 @@ spec:
           value: "default"
         - name: PORCH_REPOSITORY
           value: "focom-resources"
-        - name: KUBERNETES_BASE_URL
-          value: "https://kubernetes.default.svc"
-        - name: PORCH_HTTPS_VERIFY
-          value: "false"
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:
--- a/operators/focom-operator/docs/DEPLOYMENT.md
+++ b/operators/focom-operator/docs/DEPLOYMENT.md
@@ -284,11 +284,16 @@ NBI_STAGE=2                        # Stage 2 (Porch)
 **Optional (with defaults):**
 ```bash
 FOCOM_NAMESPACE=focom-system       # Default namespace for FOCOM resources (default: focom-system)
-KUBERNETES_BASE_URL=https://kubernetes.default.svc  # K8s API URL (auto-detected)
-TOKEN=/var/run/secrets/kubernetes.io/serviceaccount/token  # Auth token (auto-detected)
 PORCH_NAMESPACE=default            # Namespace for PackageRevisions (default: default)
 PORCH_REPOSITORY=focom-resources   # Porch repository name (default: focom-resources)
 ```
+
+In a pod, leave the API server address, token and CA unset: they are taken from
+the service account volume and the address Kubernetes advertises. Setting
+`KUBERNETES_BASE_URL` turns an auto-detected value into a fixed one, and the
+address it must then match is the only one the API server certificate is
+guaranteed to cover. `KUBERNETES_BASE_URL` must be an `https://` URL; a
+plaintext endpoint is refused, because the token is sent with every request.
 
 **Note:** The `FOCOM_NAMESPACE` environment variable sets the default namespace for all FOCOM resources (OCloud, TemplateInfo, FocomProvisioningRequest). This eliminates the need to specify namespace in API request bodies.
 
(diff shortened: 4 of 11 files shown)


thc1006:
> Hi @thc1006. Thanks for your PR.
> 
> I'm waiting for a [nephio-project](https://github.com/orgs/nephio-project/people) member to verify that this patch is reasonable to test. If it is, they should reply with `/ok-to-test` on its own line. Until that is done, I will not automatically test new commits in this PR, but the usual testing commands by org members will still work. Regular contributors should [join the org](https://github.com/orgs/nephio-project/people) to skip this step.
> 
> Once the patch is verified, the new status will be reflected by the `ok-to-test` label.
> 
> I understand the commands that are listed [here](https://go.k8s.io/bot-commands?repo=nephio-project%2Fnephio).
> 
> Details

Dear @liamfallon ,

Can u help me join to the Orgs. Bcuz I want to deal some security urgent PRs and also setup some security policy.

liamfallon (approved):

PR #1169, opened 2026-08-16
[security][o2ims] Verify Kubernetes API TLS certificates by default

thc1006:
Sibling of #1168, which fixes the same class of defect in the FOCOM operator. This one is worse, because the flag involved cannot be used correctly at all.

`controllers/utils.py` passes `HTTPS_VERIFY` as the `verify` argument of every request it makes to the Kubernetes API, and computes it as:

```python
HTTPS_VERIFY = bool(os.getenv("HTTPS_VERIFY", False))
```

`os.getenv` returns the default `False` when the variable is unset, and a string when it is set. `bool()` of any non-empty string is `True`. The result:

```console
$ python3 -c '...'
HTTPS_VERIFY=None       -> verify=False
HTTPS_VERIFY='false'    -> verify=True
HTTPS_VERIFY='False'    -> verify=True
HTTPS_VERIFY='0'        -> verify=True
HTTPS_VERIFY='true'     -> verify=True
HTTPS_VERIFY=''         -> verify=False
HTTPS_VERIFY='no'       -> verify=True
```

Verification is off by default, and every spelling of "no" turns it *on*. The only two ways to disable it are to leave the variable alone and to set it to the empty string, so the flag has never done what its name says.

### What that costs

Neither `tests/deployment/operator.yaml` nor [the deployment in nephio-project/catalog](https://github.com/nephio-project/catalog/blob/main/nephio/optional/o2ims/app/deployment.yaml) sets `HTTPS_VERIFY`, and both point the operator at `https://kubernetes.default.svc`. Every request therefore completes a TLS handshake with a peer nobody authenticated, carrying this header:

```python
HEADERS_DICT = {
    ...
    "Authorization": "Bearer {}".format(TOKEN),
}
```

The token is the operator's, and `tests/deployment/operator.yaml` binds its service account to a ClusterRole holding `create, patch` on `validatingwebhookconfigurations` and `mutatingwebhookconfigurations`, plus `create, update, patch` on `customresourcedefinitions` and full write access to `cluster.x-k8s.io` clusters. An attacker who intercepts that token does not merely read the operator's resources; they can register a mutating admission webhook and rewrite objects across the entire cluster. CWE-295 leading to CWE-522.

The one signal that would have surfaced this was switched off two lines above the flag:

```python
requests.packages.urllib3.disable_warnings()
```

That call suppresses `InsecureRequestWarning`, which urllib3 emits precisely when `verify=False`.

### What replaces it

Quoting the Kubernetes documentation on [accessing the API from a Pod](https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/):

> If available, a certificate bundle is placed into the filesystem tree of each container at `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, and should be used to verify the serving certificate of the API server.

So `tls_verify()` returns what `requests` actually wants for `verify`: that bundle when it is present, a bundle named by `KUBERNETES_CA_FILE` when one is configured, and `True` otherwise. Verification is skipped only for `UNSAFE_SKIP_TLS_VERIFY`, which is named for what it does, parsed so that anything unrecognised is false, and logged loudly when engaged. `HTTPS_VERIFY` is no longer read at all — none of its old values map onto a behaviour worth preserving — and startup says so if it is still set. `disable_warnings()` is gone.

Review closed four more paths through this, and one of them was in my own tests:

- **Plaintext was still accepted.** `requests` applies `verify` only to https, so `KUBERNETES_BASE_URL=http://...` sent the token in the clear with no opt-out required. The endpoint is now parsed once and must be an absolute `https` URL with a host, no embedded credentials, no query or fragment and a valid port; the skip flag does not make plaintext legal.
- **A missing in-cluster CA fell back to the public roots.** `True` does not mean "the system trust store" — `requests` uses the certifi bundle it ships with, which can also be redirected by `REQUESTS_CA_BUNDLE`. Falling back to it inside a pod widens the trust boundary from the cluster CA to every public authority, so an absent CA now refuses to start.
- **`isfile()` was not validation.** An empty or malformed PEM passed startup and failed on the first request. The bundle is now loaded with `ssl.create_default_context()` at startup. My original tests wrote **empty** CA files and asserted they were accepted, so they were pinning exactly the wrong behaviour — that is fixed, and those tests now use a real bundle.
- **A CA plus the skip flag was resolved silently in favour of skipping.** It is now a startup error, matching the policy #1168 adopted after its own review.

The same page also notes:

> Kubernetes does not guarantee that the API server has a valid certificate for the hostname `kubernetes.default.svc`; however, the control plane **is** expected to present a valid certificate for the hostname or IP address that `$KUBERNETES_SERVICE_HOST` represents.

The default address therefore becomes the advertised one. It used to be `http://127.0.0.1:8080`, a `kubectl proxy` whose setup has been commented out in `tests/create-cluster.sh` since before this operator was merged, and which put the bearer token on the wire in cleartext for anything listening on that port.

### Proof manifests

The truth table at the top is now a parametrized test, so the twelve spellings are pinned rather than described:

```console
$ pytest --disable-warnings -q
76 passed in 0.80s
```

Restoring the old expression in `tls_verify()` — the state this PR replaces — fails eleven of them:

```console
$ pytest --disable-warnings -q     # with tls_verify() reverted
FAILED tests/test_utils.py::test_tls_is_verified_by_default - assert False is...
FAILED tests/test_utils.py::test_in_cluster_ca_bundle_is_used_when_present - ...
FAILED tests/test_utils.py::test_kubernetes_ca_file_wins - AssertionError: as...
FAILED tests/test_utils.py::test_a_missing_ca_file_is_refused - Failed: DID N...
FAILED tests/test_utils.py::test_only_an_explicit_opt_in_skips_verification[false-False]
FAILED tests/test_utils.py::test_only_an_explicit_opt_in_skips_verification[False-False]
FAILED tests/test_utils.py::test_only_an_explicit_opt_in_skips_verification[0-False]
FAILED tests/test_utils.py::test_only_an_explicit_opt_in_skips_verification[no-False]
FAILED tests/test_utils.py::test_only_an_explicit_opt_in_skips_verification[-False]
FAILED tests/test_utils.py::test_only_an_explicit_opt_in_skips_verification[  -False]
FAILED tests/test_utils.py::test_only_an_explicit_opt_in_skips_verification[maybe-False]
11 failed, 50 passed in 0.32s
```

Worth reading the parameter names: the failures are `false`, `False`, `0`, `no`, `""`, `"  "` and `maybe` — every case that should have been verifying and was not. The `true`/`1`/`yes`/`on` cases pass under both versions, because under the old expression nothing anyone typed could have turned verification on for real.

One test asserts the plumbing rather than the policy, using the kwargs `responses` records, so the value the module computes is provably the value `requests` receives:

```python
assert responses.calls[0].request.req_kwargs["verify"] == verify
```

`flake8 controllers/utils.py` reports 18 findings on `main` and 17 here — all pre-existing `E501`s; the patch adds none and removes one long line.

One caveat on all of the above: `.prow.yaml` builds this operator's image but never runs `tox`, and the root `make unit` only walks directories containing a `go.mod`, so **nothing in `tests/` runs in CI today** and the invariant this PR pins is enforced only by whoever runs pytest locally. The runs above are local, against the versions pinned in `requirements.txt`. Wiring a Python presubmit into `.prow.yaml` would close that gap; it touches shared CI configuration that is yours to own, so it is not in this PR. The sibling #1168 had the same problem in a form I could fix inside the operator, and does.

### Effect on existing deployments

A pod needs no configuration: both the CA bundle and the address come from the cluster. The trade-off worth naming is that a cluster whose serving certificate does not cover the address the operator is pointed at will now fail the handshake instead of silently accepting it. That is the intended direction, but it is a behaviour change for anyone who was relying on the old default, and `KUBERNETES_CA_FILE` or `UNSAFE_SKIP_TLS_VERIFY` are the two ways out.

`HTTPS_VERIFY` becomes inert. The module constant is renamed `HTTPS_VERIFY` → `TLS_VERIFY` because it is no longer a boolean; nothing outside `controllers/utils.py` referenced it.

### Deliberately not in this patch

- Requests here still have no timeout, and a configuration failure is flattened into `status=False`, which `cluster_creation_request` reports as `progressing` before a 30-minute generic timeout. Both are real and both are adjacent rather than part of this boundary; raised as #1174.
- The line below the one this PR changes builds the token with `os.popen(f"cat {TOKEN}")`, which shells out with an environment variable interpolated into the command. That is a separate defect with a separate fix — reading the file directly also has to decide what to do when it is absent, which is currently masked by `cat` failing quietly — and it does not belong in a change about TLS. Raised separately as #1170.
- Moving these calls onto `kubernetes.client`, already a dependency, would make all of this the library's problem rather than ours. That is the right destination and much too large for a security fix.

### Notes for maintainers

- Companion change: `nephio/optional/o2ims/app/deployment.yaml` in nephio-project/catalog pins `KUBERNETES_BASE_URL` to `https://kubernetes.default.svc`, so catalog deployments keep that address until it is dropped there too. Verification still works on any cluster whose certificate carries that SAN, which is the usual case for kubeadm and kind; the catalog change just removes the reliance on it. Raised as nephio-project/catalog#146, whose title carries the blocker: it must not merge until this one ships in an image, because until then the operator's own default is `http://127.0.0.1:8080`.
- Documentation: `operators/o2ims-operator/README.md` gains a "Connecting to the Kubernetes API" section here. I found no reference to these variables in nephio-project/docs, so no companion PR there looks necessary.
- Release triage is yours to set. The change is confined to the o2ims operator's API client, its test deployment and its README.
- `SECURITY.md` asks that vulnerabilities not be reported through public issues. I have not opened a public issue for it. Say the word and I will route it through sig-security@lists.nephio.org before this goes further — same question as on #1168.

### DIFF
--- a/operators/o2ims-operator/README.md
+++ b/operators/o2ims-operator/README.md
@@ -105,6 +105,49 @@ kpt live init /tmp/o2ims
 kpt live apply /tmp/o2ims --reconcile-timeout=15m --output=table
 ```
 
+### Connecting to the Kubernetes API
+
+The operator verifies the API server's certificate on every request. Inside a
+pod there is nothing to configure: the CA bundle mounted at
+`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` is used, and the address
+comes from `KUBERNETES_SERVICE_HOST`, which is the one Kubernetes expects that
+certificate to be valid for.
+
+| Variable | Default | Meaning |
+|----------|---------|---------|
+| `KUBERNETES_BASE_URL` | in-cluster address | API server to talk to |
+| `KUBERNETES_CA_FILE` | in-cluster CA bundle | PEM bundle used to verify the API server |
+| `UNSAFE_SKIP_TLS_VERIFY` | `false` | Development only: skip certificate verification |
+
+Running the operator outside a cluster, point it at the API server and at the
+CA that signs its certificate:
+
+```bash
+export KUBERNETES_BASE_URL=$(kubectl config view --minify \
+  -o jsonpath='{.clusters[0].cluster.server}')
+
+# The CA is embedded in most kubeconfigs and a file path in some.
+kubectl config view --raw --minify \
+  -o jsonpath='{.clusters[0].cluster.certificate-authority-data}' \
+  | base64 -d > /tmp/cluster-ca.crt
+test -s /tmp/cluster-ca.crt || cp "$(kubectl config view --minify \
+  -o jsonpath='{.clusters[0].cluster.certificate-authority}')" /tmp/cluster-ca.crt
+export KUBERNETES_CA_FILE=/tmp/cluster-ca.crt
+
+# TOKEN is a path, not a token. Outside a pod there is no mounted one, so
+# request a short-lived credential and point at the file.
+kubectl -n o2ims create token o2ims-operator --duration=1h > /tmp/o2ims-token
+chmod 0600 /tmp/o2ims-token
+export TOKEN=/tmp/o2ims-token
+```
+
+The token expires; re-run the last three lines when it does. Inside a pod
+none of this applies, because the kubelet rotates the mounted one.
+
+`UNSAFE_SKIP_TLS_VERIFY=true` turns verification off altogether. It hands the
+service account token to an endpoint whose identity has not been checked, warns
+about it at startup, and must never be used outside development.
+
 ### Redeploying
 
 To redeploy the cluster, or to recreate the development environment, one must delete the created cluster. The Nephio mgmt cluster will be deleted automatically when running `create-cluster.sh`, but the cluster deployed by this operator has a name in the `clusterName` field. For example, it may be `edge`, thus:
--- a/operators/o2ims-operator/controllers/utils.py
+++ b/operators/o2ims-operator/controllers/utils.py
@@ -14,11 +14,12 @@
 # limitations under the License.
 ##########################################################################
 
+import logging
 import os
+import ssl
+from urllib.parse import urlsplit
 
 import requests
-
-requests.packages.urllib3.disable_warnings()
 
 TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 # Allowed values vanilla/Openshift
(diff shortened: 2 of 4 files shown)


thc1006:
Superseded by #1170, which contains this commit and the token fix as two reviewable commits against one tree.

Splitting them was a mistake on my part. The main-branch postsubmit publishes `nephio/o2ims-operator:latest` on every merge:

```
--destination=nephio/o2ims-operator:${BUILD_ID} --destination=nephio/o2ims-operator:latest
```

so merging this one alone would publish an image that verifies certificates but still builds its bearer token through `os.popen` and never re-reads it, shipping alongside a README that says the kubelet rotates it. Reviewing one final tree is also less work than reviewing two heads that each still contain the defect the other fixes.

Closing in favour of #1170. Nothing is lost: that branch is this branch plus one commit.

